PR #6, opened 2021-11-19
Change string way URL to integer way identifier

enzet:
To make specification file a bit more clear.

### DIFF
--- a/rust/osm2lanes/src/lib.rs
+++ b/rust/osm2lanes/src/lib.rs
@@ -130,7 +130,7 @@ mod tests {
     #[derive(Deserialize)]
     struct TestCase {
         /// The URL to the OSM way
-        way: String,
+        way_id: u32,
         tags: BTreeMap<String, String>,
         driving_side: DrivingSide,
         output: Vec<LaneSpec>,
@@ -153,7 +153,8 @@ mod tests {
             let actual = get_lane_specs_ltr(test.tags.clone(), &cfg);
             if actual != test.output {
                 ok = false;
-                println!("For input (example from {}):", test.way);
+                println!("For input (example from https://www.openstreetmap.org/way/{}):",
+                    test.way_id);
                 for (k, v) in test.tags {
                     println!("    {} = {}", k, v);
                 }
--- a/rust/osm2lanes/src/lib.rs
+++ b/rust/osm2lanes/src/lib.rs
@@ -130,7 +130,7 @@ mod tests {
     #[derive(Deserialize)]
     struct TestCase {
         /// The URL to the OSM way
-        way_id: u32,
+        way_id: i64,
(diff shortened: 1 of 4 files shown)


enzet:
Also, I think we should add schema version. Let's say, the previous schema was 0.1 and if this pull request is merged, the schema version will be 0.2.

rust/osm2lanes/src/lib.rs
```diff
@@ -153,7 +153,8 @@ mod tests {
             let actual = get_lane_specs_ltr(test.tags.clone(), &cfg);
             if actual != test.output {
                 ok = false;
-                println!("For input (example from {}):", test.way);
+                println!("For input (example from https://www.openstreetmap.org/way/{}):",
```
enzet, line 156: Please take a look, because I'm not sure about Rust code style and how line longer than 100 characters should be formatted.

rust/osm2lanes/src/lib.rs
```diff
@@ -130,7 +130,7 @@ mod tests {
     #[derive(Deserialize)]
     struct TestCase {
         /// The URL to the OSM way
-        way: String,
+        way_id: u32,
```
dabreegster, line 133: This should be `i64`, a signed 64-bit. I know negative IDs aren't used in OSM and I can't find a clear answer on the wiki, but I think I haven't seen people use unsigned ints. https://wiki.openstreetmap.org/wiki/Node#Structure

dabreegster (approved):
> Let's say, the previous schema was 0.1 and if this pull request is merged, the schema version will be 0.2.

We should definitely start using https://semver.org when this schema's ready to go. I think there'll be lots of churn right now, so starting to put a version on it feels early. But as long as we don't declare 1.0 or anything, not strongly opinionated

rust/osm2lanes/src/lib.rs
dabreegster: Ah good point. If you have Rust and the cargo toolchain installed locally, you can just do `cargo fmt` and it'll fix it. If you don't have it installed, feel free to leave it this way, and I'll run it when I do the next bit of work

enzet: Surprisingly it was easy to install and configure. I'll use it, thanks.

dabreegster: Yeah, the Rust build tools are super straightforward. Everything looks good to me!

enzet:
> We should definitely start using https://semver.org 

I agree.

> starting to put a version on it feels early.

OK.

rust/osm2lanes/src/lib.rs
enzet: You're right. I found that JOSM uses negative identifiers for elements not yet stored in the database.

Issue #4, opened 2021-11-17
Web interface to convert tags to lanes


dabreegster:
We could have a simple webpage hosted on github pages where you copy in an OSM way's key/values, press a button, and get the JSON lanes output. Internally it can use Javascript and call the Rust implementation, compiled to WASM.

Next steps after that could be a Leaflet or Mapbox map, where you click on a way, we query Overpass for the nearest way and its tags, and do the same thing.

And then from there, maybe consider some kind of rendering/visualization, just for a single road.

droogmic:
I had a go at this:
https://github.com/a-b-street/osm2lanes/pull/12

Helped me see that there is still quite a bit of edge-case handling missing.

enzet:
I've tried to create some sketches. What do you think?

![road_signs](https://user-images.githubusercontent.com/2399987/148714526-e58f4bfe-ebef-498b-9f56-022cea50787c.png)

dabreegster:
I think these are much more legible than Unicode, yeah! Maybe we could also print the width of each lane towards the bottom? We could also scale the width of each lane appropriately, but that stop being as legible in extreme cases

droogmic:
> I've tried to create some sketches. What do you think?
> 
> ![road_signs](https://user-images.githubusercontent.com/2399987/148714526-e58f4bfe-ebef-498b-9f56-022cea50787c.png)

are you doing these in SVG? trying to think about how to render it...

dabreegster:
If we can render each lane independently, then we could just lay them out in a row with HTML. Maybe the vertical lines in between get complicated, though... The output of the library doesn't explicitly tell you where bus and travel lanes border each other, or where the direction changes. But I guess that's easy enough to find by iterating over pairs of `LaneSpec`s...

droogmic:
I think eventually we would want a lane spec that has more information.
e.g. width
e.g. dashed, solid, double lines
e.g. line colour
e.g. bicycle symbols, and their spacing...

a lot of these will not come from OSM tags but instead need to be contributed by users as they will specific to national and regional standards.

for now we can make some assumptions in the renderer, and make some assumptions when OSM tags are missing...

enzet:
> are you doing these in SVG?

Yes, this is SVG.

> trying to think about how to render it...

I was thinking of creating SVG canvas in the HTML page and add SVG elements to it using JS. I'm not quite experienced in JS so not sure it is the right way to do this.

PR #12, opened 2022-01-09
Add Web Viewer

droogmic:
Implemented in Rust using Yew

Resolves https://github.com/a-b-street/osm2lanes/issues/4

Demo: https://droogmic.github.io/osm2lanes/

### DIFF
--- a/rust/.gitignore
+++ b/rust/.gitignore
@@ -1,0 +1,2 @@
+/target
+/dist
--- a/rust/.gitignore
+++ b/rust/.gitignore
@@ -1,2 +1,3 @@
-/target
-/dist
+dist
+target
+Cargo.lock
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -1,0 +1,14 @@
+[workspace]
+members = ["osm2lanes", "osm2lanes-web"]
+
+[profile.release]
+# less code to include into binary
+panic = 'abort'
+# optimization over all codebase ( better optimization, slower build )
+codegen-units = 1
+# optimization for size ( more aggressive )
+opt-level = 'z'
+# optimization for size 
+# opt-level = 's' 
+# link time optimization using using whole-program analysis
+lto = true
--- a/rust/Trunk.toml
+++ b/rust/Trunk.toml
@@ -1,0 +1,2 @@
+[build]
+target = "osm2lanes-web/index.html"
--- a/rust/osm2lanes-web/src/main.rs
+++ b/rust/osm2lanes-web/src/main.rs
@@ -1,0 +1,164 @@
+use std::collections::BTreeMap;
+
+use serde::{Deserialize, Serialize};
+use web_sys::HtmlInputElement;
+use yew::prelude::*;
+
+use osm2lanes::{get_lane_specs_ltr, Config, Direction, DrivingSide, LaneSpec, LaneType};
+
+// Use `wee_alloc` as the global allocator.
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+
+type ShouldRender = bool;
+
+struct WebLaneType(LaneType);
+
+impl WebLaneType {
+    /// Represents the lane type as a single character. Always picks one buffer type.
+    pub fn to_char(self) -> char {
+        match self.0 {
+            LaneType::Driving => '🚗',
+            LaneType::Biking => '🚲',
+            LaneType::Bus => '🚌',
+            LaneType::Parking => '🅿',
+            LaneType::Sidewalk => '🚶',
+            LaneType::Shoulder => 'ˢ',
+            LaneType::SharedLeftTurn => '🔃',
+            LaneType::Construction => 'x',
+            LaneType::Buffer(_) => '|',
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct State {
+    pub edit_value: String,
+    pub lanes: Vec<LaneSpec>,
+}
+
+pub enum Msg {
+    Submit(String),
+    Focus,
+}
+
+pub struct App {
+    focus_ref: NodeRef,
+    state: State,
+}
+
+fn string_to_tags(input: &str) -> Result<BTreeMap<String, String>, String> {
+    let mut map = BTreeMap::new();
+    for line in input.lines() {
+        let (key, val) = line.split_once("=").ok_or("tag must be = separated")?;
+        map.insert(key.to_owned(), val.to_owned());
+    }
+    Ok(map)
+}
+
+const CFG: Config = Config {
+    driving_side: DrivingSide::Right,
+    inferred_sidewalks: true,
+};
+
+impl Component for App {
+    type Message = Msg;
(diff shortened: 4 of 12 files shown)


enzet:
Wow! This is just awesome! Thank you for the contribution!

I think, it's better to ask @dabreegster to review since I'm not experienced in Rust.

I tried to draw some icons for different types of lanes. Maybe we can use it in the future.

rust/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[workspace]
+members = ["osm2lanes", "osm2lanes-web"]
+
+[profile.release]
+# less code to include into binary
```
dabreegster, line 5: I'm curious, are these best practices for yew projects? Performance is definitely not a concern at all for something like this project. I'm fine with all of these, just curious about the origin. FWIW I usually default to

```
[profile.dev.package."*"]
opt-level = 3
```

so even when you compile in debug mode, all the external dependencies are in release mode

rust/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
+/dist
```
dabreegster, line 2: 1) The `/dist` here didn't work for me; I wound up with `osm2lanes-web/dist/` showing up as untracked
2) I like specifying the Rust-specific ignores here; would you mind removing `target` from `../.gitignore` and moving `Cargo.lock` here please?

rust/Trunk.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+target = "osm2lanes-web/index.html"
```
dabreegster, line 2: I've never used trunk before, but wow, it's fantastic! I want to switch another huge project to using it too

rust/osm2lanes-web/src/main.rs
```diff
@@ -0,0 +1,164 @@
+use std::collections::BTreeMap;
+
+use serde::{Deserialize, Serialize};
+use web_sys::HtmlInputElement;
+use yew::prelude::*;
+
+use osm2lanes::{get_lane_specs_ltr, Config, Direction, DrivingSide, LaneSpec, LaneType};
+
+// Use `wee_alloc` as the global allocator.
```
dabreegster, line 9: Is this a standard yew thing? Is there a reason to override the allocator for wasm? (I've only ever used `wasm-bindgen` before and never had to touch it)

rust/osm2lanes-web/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+
+type ShouldRender = bool;
+
+struct WebLaneType(LaneType);
+
+impl WebLaneType {
+    /// Represents the lane type as a single character. Always picks one buffer type.
+    pub fn to_char(self) -> char {
+        match self.0 {
+            LaneType::Driving => '🚗',
```
dabreegster, line 21: Unicode is awesome :D

rust/osm2lanes-web/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+    let mut map = BTreeMap::new();
+    for line in input.lines() {
+        let (key, val) = line.split_once("=").ok_or("tag must be = separated")?;
+        map.insert(key.to_owned(), val.to_owned());
+    }
+    Ok(map)
+}
+
+const CFG: Config = Config {
+    driving_side: DrivingSide::Right,
```
dabreegster, line 60: A possible next step (doesn't have to be in this PR) could be to add a checkbox or toggle for this

rust/release.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# To be run by someone with push rights to gh-pages
+# TODO: automate this...
```
dabreegster, line 4: I reckon we could add a Github actions workflow this later

dabreegster (commented):
Wow, thank you for the PR! This is fantastic -- simple and usable, and makes it much more convenient to just try out the converter so far.

> Helped me see that there is still quite a bit of edge-case handling missing.

Indeed, that's one reason we started this project, to get OSM mappers interested in helping with some of the edge cases. :)  What kind of stuff hasn't been translating well? The biggest missing features I know about are handling per-lane widths.

dabreegster:
Some next step ideas:
- show lane width as well
- Have a text box where you can copy a URL like https://www.openstreetmap.org/way/8363012. We can query Overpass to grab the tags

rust/.gitignore
droogmic: - [ ] gitignore fixes and imprrovements

rust/osm2lanes-web/src/main.rs
droogmic: > wee_alloc is a tiny allocator that is much smaller than the allocator that is normally used in Rust binaries. Replacing the default allocator with this one will result in smaller Wasm file sizes, at the expense of speed and memory overhead.

> The slower speed and memory overhead are minor in comparison to the size gains made by not including the default allocator. This smaller file size means that your page will load faster, and so it is generally recommended that you use this allocator over the default, unless your app is doing some allocation-heavy work.

droogmic: https://yew.rs/docs/next/advanced-topics/optimizations

rust/osm2lanes-web/src/main.rs
droogmic: yes, a next step please :)

rust/release.sh
droogmic: yeah, I tried in the past, and struggled, ill let someone else with more experience with github actions try.

droogmic:
> Some next step ideas:
> 
>     * show lane width as well
> 
>     * Have a text box where you can copy a URL like https://www.openstreetmap.org/way/8363012. We can query Overpass to grab the tags

agreed,

For lane width, I think a canvas renderer would be cool.

For url: something like https://rustwasm.github.io/wasm-bindgen/examples/fetch.html to fetch from overpass.

droogmic:
> Wow, thank you for the PR! This is fantastic -- simple and usable, and makes it much more convenient to just try out the converter so far.
> 
> > Helped me see that there is still quite a bit of edge-case handling missing.
> 
> Indeed, that's one reason we started this project, to get OSM mappers interested in helping with some of the edge cases. :) What kind of stuff hasn't been translating well? The biggest missing features I know about are handling per-lane widths.

just in the example I link, I don't think taxi and bus lanes are correct, and there are more total lanes than lanes says.

droogmic:
I think I am happy with this as an MVP if it works for you, then I can iterate further with smaller PRs.

enzet:
> just in the example I link, I don't think taxi and bus lanes are correct, and there are more total lanes than lanes says.

We decided to use test-driven development. Could you please add test or just write what output we have to generate for this example? I'll try to support it.

droogmic:
> > just in the example I link, I don't think taxi and bus lanes are correct, and there are more total lanes than lanes says.
> 
> We decided to use test-driven development. Could you please add test or just write what output we have to generate for this example? I'll try to support it.

oh sure, I was going to make separate PRs for all of this, dw ;)

dabreegster:
LGTM, thank you for the PR!

Issue #3, opened 2021-11-16
Process for starting a schema


dabreegster:
# First there's the question of how we want to iterate on the schema.

1) We could get test parity with the Rust implementation as it exists currently, then iterate on the schema and keep all 3 implementations in sync.
2) We could iterate on the schema first with one implementation (Python, unless I get my Kotlin dev environment set up). Then once the dust has settled a bit, make the other 2 implementations match.

I'd vote for the second, since it seems less work.

# How to express the schema

Should we use https://json-schema.org or switch to protocol buffers or similar?

# Next steps with the schema

First thing is the set of lane types, and how we want to add details to some of them. For instance, we could have `parallel_parking` and `diagonal_parking`... or `type = parking, parking_type = parallel`. If we used protos, we could be more precise and show that the `parking_type` field is only valid when `type = parking` by using [oneof](https://developers.google.com/protocol-buffers/docs/proto3#oneof). Any preferences?

The next major question on my mind is whether to represent a bus/transit-only lane, or instead list access/usage restrictions. We _could_ just say "travel lane" with some subset of [motor vehicle, high-occupancy vehicle, public transit, taxi] allowed.

I also don't want to get too bogged down in deciding things that're pretty quick to switch in the code, but I'm not convinced blindly following A/B Street's current schema is the right choice.

enzet:
> I'd vote for the second, since it seems less work.

But I think the first one looks more consistent, doesn't it?

> Should we use https://json-schema.org

Looks good to me!

I think we should start with `type=parking`, `parking_type=parallel` and “travel lane” (or maybe “traffic lane”, anyway current `driveway` should be changed) and then see how it works.

dabreegster:
> But I think the first one looks more consistent, doesn't it?

What do you mean? If we go back and forth between different schemas, we might spend extra effort keeping all the languages in sync. How about before a major update to the schema during this prototyping phase, we send a quick draft PR and double check it's the right direction, then update all the implementations?

> I think we should start with type=parking, parking_type=parallel and “travel lane” (or maybe “traffic lane”

Sounds good. The `oneof` stuff may be overkill, and until we know all the weird edge cases we might want to represent, the extra constraint might not be necessary.

An initial sketch of lane types...

- travel lane
- parking (parallel, perpendicular, diagonal)
- sidewalk
[...]

PR #13, opened 2022-01-10
Add Spec

droogmic:
Add JSON Spec
Valdate JSON spec in tests
Accidentally reformatted the tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,9 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install pytest
+        python -m pip install --requirement requirements.txt
+    - name: Test the schema
+      run: python data/test_spec.py
     - name: Test the Python implementation
       run: pytest -v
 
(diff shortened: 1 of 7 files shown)


droogmic:
I have read https://github.com/a-b-street/osm2lanes/issues/3, and I don't really know what to do, but encoding what already exists doesn't seem like a terrible idea...

dabreegster:
Looks good to me -- we never really landed on any decision about how to express the schema, or about how it should handle all the nuance mentioned in the issue. But I agree that encoding what we're doing right now is a solid step forwards. Looks good to go, @enzet?

.github/workflows/test.yml
```diff
@@ -21,7 +21,9 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install pytest
+        python -m pip install --requirement requirements.txt
```
enzet, line 24: Great! I should have added `requirements.txt` in the first place.

enzet:
Sorry, my bad. It turned out reviews aren't visible until I submit them. :facepalm:

PR #15, opened 2022-01-10
Demo Web Canvas with Piet

droogmic:
I can draw some pretty lanes this way... but I just chose the first library I could find.

### DIFF
--- a/rust/osm2lanes-web/Cargo.toml
+++ b/rust/osm2lanes-web/Cargo.toml
@@ -4,16 +4,19 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-osm2lanes = { path = "../osm2lanes" }
 console_log = { version = "0.2", optional = true, features = ["color"] }
 log = "0.4"
+osm2lanes = { path = "../osm2lanes" }
+piet = { version = "0.5", features = ["samples"] }
+piet-web = "0.5"
+serde = { version = "1", features = ["derive"] }
+wasm-bindgen = "0.2"
 wee_alloc = { version = "0.4", optional = true }
 yew = "0.19"
-serde = { version = "1", features = ["derive"] }
 
 [dependencies.web-sys]
 version = "0.3"
-features = ["HtmlInputElement"]
+features = ["HtmlInputElement", "HtmlCanvasElement"]
 
 [features]
 default = ["console_log", "wee_alloc"]
--- a/rust/osm2lanes-web/assets/css/main.css
+++ b/rust/osm2lanes-web/assets/css/main.css
@@ -44,3 +44,8 @@ button:active, input:active, .button:active {
   align-self: center;
   flex-basis: 2em;
 }
+
+canvas {
+  width: 960px;
+  height: 480px;
+}
--- a/rust/osm2lanes-web/src/draw.rs
+++ b/rust/osm2lanes-web/src/draw.rs
@@ -1,0 +1,175 @@
+use piet::{
+    kurbo::Line, kurbo::Point, kurbo::Rect, Color, Error, FontFamily, RenderContext, Text,
+    TextAttribute, TextLayoutBuilder,
+};
+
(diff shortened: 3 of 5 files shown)


droogmic:
I will consider it done when it looks vaguely like what @enzet was showing in his graphics...

image

dabreegster:
Hmm, @enzet's icons are SVG, so we'll want easy support for loading + resizing + placing one of those. It looks like drawing an SVG to web canvas works, though a indirectly -- https://reference.codeproject.com/dom/canvas_api/drawing_dom_objects_into_a_canvas. I am not sure this is the most "modern" way to do this. The _really_ indirect way would be to use https://github.com/RazrFalcon/resvg to turn the SVG into paths and transform that into canvas instructions, but that's reinventing part of the browser.

I didn't see examples or docs related to loading SVG images in `piet`. I guess my question is, what's going to be the simplest way to render the lane cards in the browser? And will we want interactivity (clicking the lanes, maybe dragging to reorder them for example) anytime soon at all?

droogmic:
> Hmm, @enzet's icons are SVG, so we'll want easy support for loading + resizing + placing one of those. It looks like drawing an SVG to web canvas works, though a indirectly -- https://reference.codeproject.com/dom/canvas_api/drawing_dom_objects_into_a_canvas. I am not sure this is the most "modern" way to do this. The _really_ indirect way would be to use https://github.com/RazrFalcon/resvg to turn the SVG into paths and transform that into canvas instructions, but that's reinventing part of the browser.
> 
> I didn't see examples or docs related to loading SVG images in `piet`. I guess my question is, what's going to be the simplest way to render the lane cards in the browser? And will we want interactivity (clicking the lanes, maybe dragging to reorder them for example) anytime soon at all?

indeed, I don't think we will find a perfect solution.
I would also prefer to interop with streetmix before trying to recreate the drag and drop functionality.

SVG is notoriously bloated, and I am not convinced I want to be the one to implement generating SVGs and then horizontally aligning them on the page... but by all means if someone else wants to try.

If you don't mind I think I will try to finish this MVP canvas solution (I like how simple it is), and later we can add an SVG part as well...

dabreegster:
I'm always in favor of making incremental progress, so yeah, please do carry forward with this!

Just cause I'm not sure I've written out my rough plan: I'd personally like to wind up with two web apps.

The first makes it really easy to see the current lane tagging of an OSM way -- this PR directly builds towards that. The simple version lets you pick a way using a normal slippy map with existing tiles (or copy the way's URL, or for now even copy its tags manually). The more ambitious version renders the lanes directly on the slippy map. The second is much harder, because we have to deal with short road segments, [intersection geometry](https://a-b-street.github.io/docs/tech/map/geometry/index.html), and inferred road width causing the road to overlap with other things in the map.

The second step is to make it much easier to edit the lane tagging in OSM. Part of that is the UI -- something Streetmix-like is likely a clear winner. A/B Street also has an implementation of this UI, but it's currently quite limited in the types of lanes it can represent. And when we add more detail to the schema to capture lines between lanes and markings on the lanes, that'll take some UI design to not overwhelm people with options.

Another aspect of the editor is how to actually update tags from the list of `LaneSpecs`. That's probably non-trivial, unless we clobber existing lane tags and write them in one consistent way.

And finally there's the question of how to best expose this lane editor UI to people. If it wound up in iD, that's probably the best way to reach lots of mappers, but it's going to be much more logistically and technically difficult to squeeze a big new feature into one of the main editors.

droogmic:
Long road ahead ;)

The one that seems most tractable for my skills is the roundtrip lanes2osm2lanes, as long as I don't have to do all the implementation in all 3 languages...

I found https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage, which might make stitching SVGs together on a canvas quite easy.

droogmic:
Example republished: https://droogmic.github.io/osm2lanes/

dabreegster:
LGTM, thanks for the PR! I'll get github actions wired up "soon"...

PR #17, opened 2022-01-11
Rust: Refactor Tags Struct

droogmic:
Regroup Tags to be part of the public interface, as it appears to be widely used.

### DIFF
--- a/rust/osm2lanes/src/lib.rs
+++ b/rust/osm2lanes/src/lib.rs
@@ -147,11 +147,17 @@ impl LanePrintable for Direction {
 }
 
 /// Internal convenience functions around a string->string map
-struct Tags(BTreeMap<String, String>);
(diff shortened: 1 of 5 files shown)


rust/osm2lanes/src/lib.rs
```diff
@@ -147,13 +147,19 @@ impl LanePrintable for Direction {
 }
 
 /// Internal convenience functions around a string->string map
```
dabreegster, line 149:
```suggestion
/// A map from string keys to string values. This makes copies of strings for 
/// convenience; don't use in performance sensitive contexts.
```

dabreegster (approved):
LGTM! Working with strings in Rust can get unergonomic, and this is a performance-neglecting way around that. Fine to expose.

PR #21, opened 2022-01-12
Rust: Add Warnings

droogmic:
OSM to lane transformation cannot always be perfect

There will often be cases where we can or need to make assumptions, these should be reported as warnings.
There will sometimes be conflicting tags, these should be reported as errors.

The interface gets updated with these types.


dabreegster:
Nicely structured -- both warnings and errors may be relevant to the caller.

I guess one of the upfront errors might be missing `highway` entirely, or having something not representing a road. My version of that so far is https://github.com/a-b-street/abstreet/blob/c6ec390d0e2a01d147a46fd6337522f1e3ba00c2/convert_osm/src/extract.rs#L355, but it makes lots of assumptions that we should question heavily for this project.

PR #22, opened 2022-01-13
Rewrite Spec

droogmic:
I think the current spec is a bit brittle if we want to be truly agnostic to our clients

Adds:
designated vehicle - to help renderers
use - to inform on access restrictions by use case.
width: to describe width of the lane.

Open issue: Something I have been mulling over in my head is whether we shouldn't make all lane separators be just a variant of lane. Because what is the difference between a painted line, 2 painted lines, 2 painted lines with some more space in between, 2 painted lines with hashmarks inbetween, bollards, etc. IMO it will become important to describe in full detail what kind of vehicles can cross, e.g. for overtaking, and how wide the space is.


README.md
```diff
@@ -30,18 +30,108 @@ Input JSON file with road OpenStreetMap tags:
...
     {"type": "sidewalk", "direction": "forward"}
 ]
 ```
 
+Lane Definition
+---------------
+
+It is important to determine a strict definition for a lane.
+An OpenStreetMap way, usually a `highway=*`, can in some cases be divided into lanes,
+where each lane extends from the start to the end of the way.
```
dabreegster, line 51: "where each lane extends from the start to the end of the way."

I've been puzzling over this one for a long time in the context of street parking. A simple issue is that a parking "lane" is regularly interrupted for driveways. The OSM wiki (I could dig up a reference) does _not_ recommend splitting the way constantly to deal with this, and in practice nobody does, which I think is good. It may be something we want to clarify in the spec -- in this case, a parking lane's use is split between storing vehicles and keeping driveways clear.

It gets more complicated with "pocket parking" and "bulb-outs". One image I found is https://www.susdrain.org/images/derbyshirestreet_fig1.jpg, but I could go take better pictures. Many streets change width frequently on one block.

So far, I think we've been choosing to sacrifice this level of detail, because mapping it is hard in practice and it greatly complicates any renderers/tools built on top.

Just wanted to raise this issue generally, see if you have thoughts on a better representation

README.md
```diff
@@ -30,18 +30,108 @@ Input JSON file with road OpenStreetMap tags:
...
+Note:
+
+- there is are shoulders where pedestrians could walk in both directions.
+- this is in the UK, so traffic drives on the left.
+- the direction is the predominant direction,
+  overtaking vehicles could travel in the opposite direction.
+
+### Example 3 ###
+
+A narrow road with one lane.
```
dabreegster, line 100: I like modeling this as one travel lane with `direction=both`. I hereby regret my early choice of creating two skinny one-direction lanes to handle these.

README.md
```diff
@@ -30,18 +30,108 @@ Input JSON file with road OpenStreetMap tags:
...
+
+A road with no marked lanes.
+
+![No Lanes India](https://upload.wikimedia.org/wikipedia/commons/thumb/5/5a/Bijupara-Khalari_Road_-_Jharkhand_1648.JPG/1920px-Bijupara-Khalari_Road_-_Jharkhand_1648.JPG)
+
+![No Lanes Delhi](https://upload.wikimedia.org/wikipedia/commons/thumb/a/a8/Ratan_Lal_Market%2C_Kaseru_Walan%2C_Paharganj%2C_New_Delhi%2C_Delhi%2C_India_-_panoramio_%281%29.jpg/1280px-Ratan_Lal_Market%2C_Kaseru_Walan%2C_Paharganj%2C_New_Delhi%2C_Delhi%2C_India_-_panoramio_%281%29.jpg)
+
+![No Lanes Dutch](https://upload.wikimedia.org/wikipedia/commons/e/e8/Fietsstraat.jpg)
+
+Option 1: Mark the entire road as a single lane in both directions
```
dabreegster, line 124: This gets my vote -- it's like the narrow lane example above, except not so narrow. Even on the narrow lane, two opposing vehicles would _probably_ stick to the usual driving direction by default, depending on the exact situation and geometry.

For renderers, one big unmarked bidi lane seems easy to handle. Likewise for routing. Things like traffic simulators have to make some choice about how much detail to model, and internally, it'd be fine for them to treat this as two one-direction lanes with conventional directions. I think that modeling choice really will vary, and this library shouldn't get opinionated about it

dabreegster:
Lane separators as lanes -- hmm, yeah, that _is_ sounding like a nice representation. That would also handle representing a dual carriageway with some large grass / concrete median as one set of lanes.

I wonder if that leads us to a per-lane property about whether different people/vehicles can move left or right. For example, moving between a cycle lane through some plastic bollards spaced apart to a travel lane may be possible for cyclists (with some sort of penalty), impossible for cars. Or when a road has a center-line marked as "no passing", we can represent that, vs "over-taking in the opposite direction allowed here"

Pulling in @enzet and @kfarr for thoughts too

README.md
droogmic: hmm, as with everything in OSM, we should just support some incremental addition of detail:

1. Some way
2. parking=yes
3. parking=parallel_excluding_driveways (or something) and any other common arrangements
4. micromapping the parking spots as areas, no longer attached to the way? This will add complexity when we need to stitch those back together with the roads, so I suspect this will only come after we have a good way of doing road areas...

README.md
droogmic: agreed, I will select option 1 but leave the considered options there so it is recorded...

kfarr:
> Lane separators as lanes
@dabreegster yes as a practical matter this is how we have handled rendering lane separators -- we first parse segments from Streetmix, and then we insert "0-width" segments with different variants such as "double yellow", etc. It has worked out very nicely and it also opens the door to allow editing of the 0-width segments to allow for the user to easily change striping variants (when we add planned editing features)

README.md
```diff
@@ -30,18 +30,108 @@ Input JSON file with road OpenStreetMap tags:
 }
 ```
 
-Output lane specifications from left to write:
+Output: A list of lanes from left to write according to the [specification](./data/spec-lanes.json).
+JSON formatted example:
 
 ```json
 [
-    {"type": "sidewalk", "direction": "backward"},
```
droogmic, line 37: revert

README.md
```diff
@@ -30,18 +30,108 @@ Input JSON file with road OpenStreetMap tags:
...
+    {"type": "travel_lane", "direction": "backward"},
+    {"type": "shoulder", "direction": "both"},
+]
+```
+
+![Example 2](https://upload.wikimedia.org/wikipedia/commons/f/f0/A537_Cat_and_Fiddle_Road_-_geograph.org.uk_-_175899.jpg)
+
+Note:
+
+- there is are shoulders where pedestrians could walk in both directions.
```
droogmic, line 93: there are*

droogmic:
I will rewrite this proposal asssuming any non-zero lane seperation is itself a lane, and see what it looks like.

dabreegster:
@droogmic, if it'd be helpful to have a brainstorming session by videocall with @kfarr or anyone else, let me know and I can schedule something. I think different people have been thinking about similar problems for a long time, it's awesome to see real progress happening now! Also fine with async communication like we've been doing

droogmic:
> @droogmic, if it'd be helpful to have a brainstorming session by videocall with @kfarr or anyone else, let me know and I can schedule something. I think different people have been thinking about similar problems for a long time, it's awesome to see real progress happening now! Also fine with async communication like we've been doing

I don't mind, I get enough of that with work. I'll let you schedule and chair one if you think it is necessary. For now I am happily making incremental progress :P.

droogmic:
Ok, so I now have a rewritten test set, as well as a rewritten python implementation.

@dabreegster, now is the time for a decision on whether or not to proceed (including a call or not)

README.md
```diff
@@ -12,39 +12,134 @@ Structure
...
+    {"type": "travel_lane", "direction": "forward"},
+]
 ```
 
-Output lane specifications from left to write:
+Note:
+
+- there is no sidewalk nor verge for pedestrians to walk here.
+- lane separators are listed.
+- we omit widths for simplicity. (TODO: if a guess is needed, e.g. on an average lane width, where is this guess made?)
```
dabreegster, line 57: That's one of the next things I'm interested in adding. The Berlin group behind [this map](https://supaplexosm.github.io/strassenraumkarte-neukoelln/posts/2021-12-31-micromap-update) has been adding cycleway and separator widths in OSM.

I think this guess could live in `osm2lanes`. Something like
```
width: 2.0,    // always meters, per the spec
width_source: "osm | assumption"
```

Defaults probably vary by `highway` type and region. We can think about some "profiles" in `Config`

README.md
```diff
@@ -12,39 +12,134 @@ Structure
 ---------
 
   * `data`
-    * `tests.json`—tests, initially converted from Rust file
-      `map_model/src/make/initial/lane_specs.rs` of A/B Street project.  Code is
-      under Apache-2.0 License.
+    * `tests.json`—tests, initially converted from Rust file.
```
dabreegster, line 15: Update now that this is yaml?

data/spec-lanes.json
```diff
@@ -1,28 +1,39 @@
...
-                    "construction",
-                    "cycleway",
-                    "parking_lane",
-                    "shared_left_turn",
-                    "shoulder",
-                    "sidewalk",
-                    "travel_lane"
+                "anyOf": [
+                    {
+                        "const": "travel",
```
dabreegster, line 18: So the big change is to shift from existing "categories" of common lane types like cycleways and shared left turn lanes, to a general travel lane that decomposes concerns. `designated` seems to be the access restrictions -- so a "bus lane" also usable by taxis and cyclists can be indicated directly. Is that the idea?

data/spec-lanes.json
```diff
@@ -44,6 +55,90 @@
                         "description": "Traffic does not travel in this lane."
                     }
                 ]
+            },
+            "designated": {
+                "type": "string",
+                "description": "The predominant or designated traffic type, to hint at how to depict the lane. Should not be used to determine access restrictions.",
```
dabreegster, line 61: Should not be used for access restrictions -- why not? How else do we know that a cycle lane alongside a road doesn't allow cars to squeeze in and drive through? Aside from this second sentence, this look good to me

data/spec-lanes.json
```diff
@@ -44,6 +55,90 @@
...
+                        "const": "bus"
+                    },
+                    {
+                        "const": "taxi"
+                    }
+                ]
+            },
+            "markings": {
+                "type": "array",
+                "description": "The road markings arranged from left to right.",
```
dabreegster, line 89: Is this valid _only_ for `type=separator`? If not, what's an example?

data/tests.yml
```diff
@@ -0,0 +1,570 @@
+- way_id: 159276650
```
dabreegster, line 1: The advantages of YAML seem to be
1) less "noise" for quickly writing things like output in a text editor
2) Proper comments

Were there other motivations?

dabreegster (commented):
I raised a few questions, but overall I like this and am good with it!

data/spec-lanes.json
droogmic: I would prefer to have another key: `access`, which contains a tags matching https://wiki.openstreetmap.org/wiki/Key:access#Transport_mode_restrictions for that specific lane.

A counterexample for designated=bicyclem, but cars allowed, is https://nl.wikipedia.org/wiki/Fietsstraat

data/spec-lanes.json
droogmic: - [ ] valid only for type=separator

data/tests.yml
droogmic: + easier to write by hand
-  yaml is really bad (ideally i would use https://hitchdev.com/strictyaml/why/, but support is less good)

data/spec-lanes.json
dabreegster: Got it. From a routing perspective, this sounds like the difference between "allowed" and "encouraged" -- driving down a "living street" is acceptable, but not prioritized. Cycling on a path through a park primarily meant for foot traffic is allowed, but not expected to be fast.

droogmic:
To be clear, I am not happy with how big this has become :(
necessary evil?

data/spec-lanes.json
droogmic: this is to help clients differentiate lanes by the common use cases, e.g. the options in streetmix

data/spec-lanes.json
```diff
@@ -38,13 +49,128 @@
...
+                        "const": "bus"
+                    },
+                    {
+                        "const": "psv"
                     }
                 ]
+            },
+            "access": {
+                "type": "object",
+                "description": "Access to the lane. Freeform tags matching OSM way access",
```
dabreegster, line 85: So we don't further help the consumer with an enum. Seems fine for now

kotlin/src/test/kotlin/Test.kt
```diff
@@ -37,9 +37,9 @@ internal class LaneTest {
             if (!testCase.skip) {
                 val parsed = Road(testCase.tags, testCase.driving_side).parse()
                 assertEquals(
-                    testCase.output,
+                    testCase.output.filter { it.type != LaneType.SEPARATOR },
                     parsed,
-                    "\nGot:      " + parsed + "\nExpected: " + testCase.output + "\n"
+                    "${testCase.way_id}\nGot:      ${parsed}\nExpected: ${testCase.output}\n"
```
dabreegster, line 42: The Kotlin code is still using the JSON, but now that file is deleted from git and under `.gitignore`. Is this just temporary? The goal is to move everything to using YAML?

rust/osm2lanes/src/lib.rs
```diff
@@ -124,50 +101,70 @@ pub trait LanePrintable {
...
-            LaneType::Buffer(_) => '|',
+            Self::Travel {
+                designated: LaneDesignated::Foot,
+                ..
+            } => 's',
+            Self::Travel {
+                designated: LaneDesignated::Bicycle,
+                ..
+            } => 'b',
+            Self::Travel {
```
dabreegster, line 115: Trade-off: This kind of code becomes more complex, having to match something embedded. But it buys us flexibility.

rust/osm2lanes/src/transform.rs
```diff
@@ -378,119 +412,82 @@ fn bicycle(
             ));
         }
     }
-
-    // My brain hurts. How does the above combinatorial explosion play with
```
dabreegster, line 382: Hard to follow from the JSON -> YAML diff. Did some of these broken or edge cases get implemented better, or was the test added to the ignore-list for now?

rust/osm2lanes/src/transform.rs
```diff
@@ -622,51 +616,67 @@ fn assemble_ltr(
...
+        _oneway = true;
     }
     // Pedestrian
     {
         match (
-            lanes.first().unwrap().lane_type == LaneType::Sidewalk,
-            lanes.last().unwrap().lane_type == LaneType::Sidewalk,
+            matches!(
+                lanes.first().unwrap(),
+                Lane::Travel {
```
dabreegster, line 668: Similar to the `Lane::foot()` constructor, I wonder if we need `lane.is_foot()` predicates to shorten all this boilerplate? Just an idea going forward, not sure. By making the schema more flexible, it'll mean every language implementation will grow its own helper methods like this, which might diverge over time

rust/osm2lanes/src/transform.rs
```diff
@@ -696,13 +720,45 @@ pub fn lanes_to_tags(lanes: &[LaneSpec], cfg: &Config) -> Result<Tags, LaneSpecE
...
+                        ..
+                    }
+                )
+            });
         let right_cycle_lane = lanes
             .iter()
             .rev()
-            .take_while(|lane| lane.lane_type != LaneType::Driving)
-            .find(|lane| lane.lane_type == LaneType::Biking);
+            .take_while(|lane| {
```
dabreegster, line 744: And yeah... in terms of readability, the one-liner was so much better. :\   I might add the predicates after this PR.

kotlin/src/test/kotlin/Test.kt
droogmic: I couldn't get native yaml parsing working, so I have the python dump a json :)
feel free to fix it, but im not good at the kotlin + the DSL for dependencies :P

rust/osm2lanes/src/transform.rs
droogmic: I could actually delete it without breaking anything, will make a separate PR to demosntrate :)

rust/osm2lanes/src/transform.rs
droogmic: yeah, I think it can be attributed to the increased complexity in the schema?

rust/osm2lanes/src/transform.rs
droogmic: https://github.com/a-b-street/osm2lanes/pull/33

dabreegster (approved):
Sorry for the slow review. I'm happy with the general spirit of this change, and agree we can keep iterating on things like the access tags, helper methods for Rust, YAML parser in Kotlin

droogmic:
A bit worrying merging such a big change, but I guess we can always revert if needed...
Ill merge this when I get back tonight.

dabreegster:
Nothing in production is using this yet, so merge whenever

droogmic:
On that note, I would like to make steps towards including it as a
dependency in some existing application, because it would give the feature
development more focus. Would that be abstreet or something else, and what
are the missing features to get it ready for that?

dabreegster:
> Would that be abstreet or something else, and what are the missing features to get it ready for that?

I agree we're getting past the point of this just being an experiment. One idea is taking the nice web app and loudly announcing that. It may get more use/attention once it's glued to a Streetmix-like editor, or something that can render a bunch of roads in an area. So we're not really close to that stuff yet.

It's always been my intention to cut A/B Street over to this. There's a bit of maintenance cost to working with two separate git repos, but it'll be fine. For local development when osm2lanes might affect A/B Street, I can just pin the dependency to a local copy.

In terms of what it'd take to cutover... the main missing feature is lane widths. The current implementation is pretty simple -- and bad: https://github.com/a-b-street/abstreet/blob/0367a7238c7257a30d63b7be9c368d4346572b68/map_model/src/objects/lane.rs#L495
I think we could generalize this better by encoding this in the `Config` struct somehow, and including some default "profiles." I'm sure in some countries / cities, people will contribute more details in the form of official design manuals or informal typical values.

To ease the A/B Street codebase into this change, I'll probably also preserve the current approach of bidirectional lanes being split into two one-way pieces, but that transformation can live over there, until I come up with something better in the simulation / UI layers.

Another difficulty will be handling pedestrian-focused spaces; A/B Street currently pretends those are cycleways with tiny walkable shoulders. That's a hack driven by older priorities, but yet to be cleaned up. I think also that can be papered over outside of `osm2lanes`. So in general A/B Street will probably be the most complicated consumer, burdened with some legacy decisions, and have pre/post transformations that ideally get cleaned up over time.

dabreegster:
Another idea is adding a feature to 3D Street to take an OSM way ID, download it with Overpass, run through `osm2lanes`, and transform that into 3D Street's version of lane specs. @kfarr, interest from your side in doing that soon?

kfarr:
@dabreegster quick answer yes, long answer it would help to chat to review the high-level workflow and capabilities, and ensure this all fits in line with our goals.

droogmic:
In the mean time, I'll keep working towards adding capabilities and displaying them in the web viewer

PR #23, opened 2022-01-13
Rust: Prepare for Rountrip Testing

droogmic:
Prepare the boilerplate for lanes2osm.

We roundtrip lanes->osm->lanes, and not the other way around, because we expect osm to lanes to be a many to one mapping.

### DIFF
--- a/rust/osm2lanes/src/lib.rs
+++ b/rust/osm2lanes/src/lib.rs
@@ -8,7 +8,7 @@ use std::collections::BTreeMap;
 
 use serde::{Deserialize, Serialize};
 
-pub use self::transform::get_lane_specs_ltr;
+pub use self::transform::{get_lane_specs_ltr, lanes_to_tags};
 
 mod transform;
 
@@ -218,6 +218,14 @@ mod tests {
         skip: Option<bool>,
     }
 
+    fn stringify_lane_types(lanes: &[LaneSpec]) -> String {
+        lanes.iter().map(|s| s.lane_type.as_ascii()).collect()
+    }
+
+    fn stringify_directions(lanes: &[LaneSpec]) -> String {
+        lanes.iter().map(|s| s.direction.as_utf8()).collect()
+    }
+
     #[test]
     fn test_from_data() {
         // TODO This is brittle, depends on running the test from the right directory. Use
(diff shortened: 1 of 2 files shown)


rust/osm2lanes/src/lib.rs
```diff
@@ -261,11 +269,49 @@ mod tests {
         assert!(ok);
     }
 
-    fn stringify_lane_types(lanes: &[LaneSpec]) -> String {
-        lanes.iter().map(|s| s.lane_type.as_ascii()).collect()
-    }
+    #[test]
+    fn test_roundtrip() {
+        // TODO This is brittle, depends on running the test from the right directory. Use
```
dabreegster, line 274: You can remove this and the copy -- I can't find any mention of how it works, but the test runner seems to get this right and set the working directory consistently, no matter where you run `cargo test`. https://stackoverflow.com/questions/30003921/how-can-i-locate-resources-for-testing-with-cargo asked this a while ago, I'm guessing something changed since then

rust/osm2lanes/src/transform.rs
```diff
@@ -485,3 +485,38 @@ fn osm_separation_type(x: &str) -> Option<BufferType> {
         _ => None,
     }
 }
+
+pub fn lanes_to_tags(lanes: &[LaneSpec], _cfg: &Config) -> Result<Tags, LaneSpecError> {
+    let mut tags = std::collections::BTreeMap::new();
+    tags.insert("highway".to_owned(), "yes".to_owned()); // TODO, what?
```
dabreegster, line 491: Now there's a good question. :)   We don't use all the tags originally, so I think emitting a subset of them is reasonable. Except we _do_ use this one, only sometimes. I wonder if we're forced to stash it?

rust/osm2lanes/src/transform.rs
```diff
@@ -485,3 +485,38 @@ fn osm_separation_type(x: &str) -> Option<BufferType> {
...
+    }
+    if lanes
+        .iter()
+        .filter(|lane| lane.lane_type == LaneType::Driving)
+        .all(|lane| lane.direction == Direction::Forward)
+    {
+        tags.insert("oneway".to_owned(), "yes".to_owned());
+    }
+    if lanes.first().unwrap().lane_type == LaneType::Sidewalk
+        && lanes.first().unwrap().lane_type == LaneType::Sidewalk
```
dabreegster, line 507: `last`!

dabreegster (changes requested):

rust/osm2lanes/src/transform.rs
droogmic: *but the tests passed*

TDD taken to the extreme has its limits :P

PR #24, opened 2022-01-13
Rust: Refactor Bicycle - First Pass

droogmic:
Attempt to cleanup bicycle lane processing.
Ended up making thinigs worse...

While doing this I determined I want better tag combination semantics. Passing around Option<Enum> values is not fun.
I also needed to disable some tests pending bidirectional cyclepath support.

Please advise on how incremental you want me to do this...

All skipped tests are due to bidirectional or LHD cyclepaths.

### DIFF
--- a/rust/osm2lanes/src/lib.rs
+++ b/rust/osm2lanes/src/lib.rs
@@ -164,7 +164,9 @@ impl LanePrintable for Direction {
 
 /// A map from string keys to string values. This makes copies of strings for
 /// convenience; don't use in performance sensitive contexts.
-// TODO: Rationale for BTreeMap iso HashMap
+// BTreeMap chosen for deterministic serialization.
+// We often need to compare output directly, so cannot tolerate reordering
+// TODO: fix this in the serialization by having the keys sorted.
 #[derive(Clone, Deserialize, Default)]
 pub struct Tags(BTreeMap<String, String>);
 
--- a/rust/osm2lanes/src/lib.rs
+++ b/rust/osm2lanes/src/lib.rs
@@ -261,10 +261,11 @@ mod tests {
     }
 
     impl DrivingSide {
+        /// Three-letter abbreviation
         const fn to_tla(&self) -> &'static str {
             match self {
-                Self::Right => "RHD",
-                Self::Left => "LHD",
+                Self::Right => "RHT",
+                Self::Left => "LHT",
             }
         }
     }
--- a/rust/osm2lanes/src/lib.rs
+++ b/rust/osm2lanes/src/lib.rs
@@ -8,9 +8,8 @@ use std::collections::BTreeMap;
 
 use serde::{Deserialize, Serialize};
 
-pub use self::transform::{get_lane_specs_ltr, lanes_to_tags};
-
 mod transform;
+pub use self::transform::{get_lane_specs_ltr, get_lane_specs_ltr_with_warnings, lanes_to_tags};
 
 /// A single lane
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
@@ -100,6 +99,15 @@ pub enum DrivingSide {
     Left,
 }
 
+impl DrivingSide {
+    pub fn opposite(&self) -> Self {
+        match self {
+            Self::Right => Self::Left,
+            Self::Left => Self::Right,
+        }
+    }
+}
+
 /// Display lane detail as printable characters
 pub trait LanePrintable {
     fn as_ascii(&self) -> char;
@@ -156,7 +164,8 @@ impl LanePrintable for Direction {
 
 /// A map from string keys to string values. This makes copies of strings for
 /// convenience; don't use in performance sensitive contexts.
-#[derive(Clone, Deserialize)]
+// TODO: Rationale for BTreeMap iso HashMap
+#[derive(Clone, Deserialize, Default)]
 pub struct Tags(BTreeMap<String, String>);
 
 impl Tags {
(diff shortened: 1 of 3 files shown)


droogmic:
I made a mistake, all RHD should be RHT, and all LHD should be LHT

- [ ] all RHD should be RHT, and all LHD should be LHT

rust/osm2lanes/src/lib.rs
```diff
@@ -148,7 +156,8 @@ impl LanePrintable for Direction {
 
 /// A map from string keys to string values. This makes copies of strings for
 /// convenience; don't use in performance sensitive contexts.
-#[derive(Clone, Deserialize)]
+// TODO: Rationale for BTreeMap iso HashMap
```
dabreegster, line 159: Ah, it's just for deterministic serialization. In lots of projects I'm making sure some pipeline doesn't behaviorally change by comparing output directly, so I try and trim f64's and keep the order of output the same. It may be convenient to have that determinism here too. But maybe also there's better ways of doing it, like having a custom serde serializer that does the sorting last minute

rust/osm2lanes/src/lib.rs
```diff
@@ -216,6 +225,36 @@ mod tests {
...
+                println!("        Comment: {}", comment);
+            }
+            for (k, v) in self.tags.map() {
+                println!("    {} = {}", k, v);
+            }
+        }
+    }
+
+    impl DrivingSide {
+        const fn to_tla(&self) -> &'static str {
```
dabreegster, line 252: I'm a bit lost on the acronyms.. LHD is left handed driving? TLA is?

rust/osm2lanes/src/transform.rs
```diff
@@ -70,8 +110,7 @@ fn non_motorized(tags: &Tags, cfg: &Config) -> Option<LaneSpecResult> {
     // types, assume bikes are allowed, except for footways, where they must be explicitly
     // allowed.
     if tags.is("bicycle", "no")
-        || (tags.highway_is("footway")
-            && !tags.is_any("bicycle", &["designated", "yes", "dismount"]))
+        || (tags.highway_is("footway") && !tags.is_any("bicycle", &["designated", "yes"]))
```
dabreegster, line 113: What happened to dismount?

rust/osm2lanes/src/transform.rs
```diff
@@ -196,65 +233,157 @@ fn bicycle(
...
+                tags: tags.subset(&[CYCLEWAY]),
+            });
+            backward_side.push(LaneSpec::backward(LaneType::Biking));
         }
-        if tags.is("cycleway:left", "opposite_lane") || tags.is("cycleway", "opposite_lane") {
-            if cfg.driving_side == DrivingSide::Right {
-                backward_side.push(LaneSpec::backward(LaneType::Biking));
-            } else {
-                forward_side.push(LaneSpec::forward(LaneType::Biking));
+        // cycleway:FORWARD=*
```
dabreegster, line 315: I am still a bit lost (it feels like there are so many combinations to consider), but I think this is becoming more thorough + organized than the original, so thumbs up from me

rust/osm2lanes/src/lib.rs
droogmic: any rationale is fine for now, so that it can be changed confidently later when the rationale changes :)

rust/osm2lanes/src/lib.rs
droogmic: TLA = three-letter acronym
it's funnier to leave it abbreviated, but I will expand it out :P

rust/osm2lanes/src/transform.rs
droogmic: I wasn't convinced it was correct...

```
highway=footway
bicycle=dismount
```

That's a sidewalk, or am I missing something?

rust/osm2lanes/src/lib.rs
dabreegster: Oh wow there's some irony here...

rust/osm2lanes/src/transform.rs
droogmic: Agreed, it's still not great. I think I want to iterate on the spec before merging something like this. Hopefully then it gets better.

rust/osm2lanes/src/transform.rs
dabreegster: Ahh nope, you're right. Originally I was hacking around stuff in A/B Street by treating this as a cycle lane, to make sure stuff was connected. That hack makes no sense here.

https://taginfo.openstreetmap.org/tags/bicycle=dismount#combinations
For reference, there are some ways that use both these tags -- so they kind of indicate bike use is explicitly allowed, but only with dismounting. Not sure how to fit that in this spec yet.

droogmic: I think it's a bit more useful for routers, to give the way a low speed but high desirability for cyclists?

dabreegster: Yeah, makes sense. I've experienced a few of these around Vauxhall in London particularly, where the intended design is probably to stop and act like a pedestrian briefly.

I wonder how to reflect this in the spec. Maybe instead of access per vehicle type being binary, some kind of "allowed, but with a penalty"?

droogmic: depends what the libraries clients are interested in.
I wonder if most just don't care, and would be more happy pretending it is a pedestrian way.
In the draft of the spec I have, I use "designated" to give a hint on how to draw the lane.

I would later want to extend the spec with more yes/no/designated/dismount information per lane.
I guess that is what you need for abstreet? essentially lane by lane routing?
so then have access tags returned for a lane mimic the access tags for a one lane highway?

rust/osm2lanes/src/lib.rs
enzet: Sorry for the off topic, but this is funniest smart joke I ever seen on the GitHub. Hope you don't mind if I screen it.

rust/osm2lanes/src/transform.rs
dabreegster: > I wonder if most just don't care, and would be more happy pretending it is a pedestrian way.

On second thought, this makes more sense to me. No need to complicate things before a particular client needs it.

rust/osm2lanes/src/transform.rs
```diff
@@ -4,13 +4,112 @@ use serde::{Deserialize, Serialize};
...
+            DrivingSide::Left => Self::from("left"),
+        }
     }
-    fn highway_is_any(&self, values: &[&str]) -> bool {
-        self.is_any(Self::HIGHWAY, values)
+}
+
+impl DrivingSide {
+    fn tag(&self) -> TagKey {
+        (*self).into()
```
dabreegster, line 112: Style nit: Why not making all the tiny enums be `Copy` and take `self` here?

dabreegster (approved):
Nice! Seeing the context of `TagKey`, I agree it helps.

I'll merge this now, feel free to make enums `Copy` if you like that style in a later PR. I also gave you write permission on the repo; feel free to push small commits without any review.

And I automated the web deployment: https://a-b-street.github.io/osm2lanes/
It's pretty easy once you find a small example to copy from and stop messing up relative directories. Unfortunately there's no real way to test github action config locally, without just pushing a bunch of attempts in another branch...

PR #25, opened 2022-01-14
Rust: Proposal on tag key concatenation

droogmic:
Still trying to make a more idiomatic representation for colon separated keys.

Will probably then repeat this for pipe and semi-colon separated values if we are happy with it.

### DIFF
--- a/rust/osm2lanes/src/transform.rs
+++ b/rust/osm2lanes/src/transform.rs
@@ -4,13 +4,60 @@ use serde::{Deserialize, Serialize};
 
 use crate::{BufferType, Config, Direction, DrivingSide, LaneSpec, LaneType, Tags};
 
+enum TagKey {
+    Const(&'static str),
+    Str(String),
+}
+
+impl TagKey {
+    const fn from(string: &'static str) -> Self {
+        TagKey::Const(string)
+    }
+    fn as_str(&self) -> &str {
+        match self {
+            Self::Const(v) => v,
+            Self::Str(v) => v.as_str(),
+        }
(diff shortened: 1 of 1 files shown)


rust/osm2lanes/src/transform.rs
```diff
@@ -145,14 +192,14 @@ fn bus(
     forward_side: &mut [LaneSpec],
     backward_side: &mut [LaneSpec],
 ) {
-    let fwd_bus_spec = if let Some(s) = tags.get("bus:lanes:forward") {
+    let fwd_bus_spec = if let Some(s) = tags.key_get(TagKey::from("bus") + "lanes" + "forward") {
```
dabreegster, line 195: I'm not sure I understand the problem being addressed? I was kind of expecting some abstraction that handles separators in the values (like `turn:lanes=none|none|through;slight_right`). I guess some keys do get complex, like `cycleway:left:width`. But then I'd expect there to be something breaking apart this string into pieces.

dabreegster: Ah, from the PR description, maybe I follow a bit better.. are you just trying to conveniently handle a short vector of strings and deal with `&'static str` vs `String`?

droogmic: yeah, the idea came from the other PR where I needed the `:left:` and `:right:` parts to be programmatic.
I am not convinced though, so I'll leave it on ice until I can better justify needing the complexity :/

droogmic:
merged into PR#24

PR #26, opened 2022-01-15
Add Support For Lane Direction Both

droogmic:
Also: add information to test data including mapillary and comments for improvement


dabreegster:
That's lots of work grabbing Mapillary references, but super useful -- thanks!

droogmic:
> That's lots of work grabbing Mapillary references, but super useful -- thanks!

yeah, but I think in the long run it will help having images for all real-life examples.

PR #27, opened 2022-01-15
Rust: Finish Lane to OSM Tags

droogmic:
Next up: upgrade the web version with this.

### DIFF
--- a/rust/osm2lanes/src/transform.rs
+++ b/rust/osm2lanes/src/transform.rs
@@ -646,7 +646,9 @@ pub fn lanes_to_tags(lanes: &[LaneSpec], _cfg: &Config) -> Result<Tags, LaneSpec
         let lane_count = lanes
             .iter()
             .filter(|lane| {
-                lane.lane_type == LaneType::Driving || lane.lane_type == LaneType::SharedLeftTurn || lane.lane_type == LaneType::Bus
+                lane.lane_type == LaneType::Driving
+                    || lane.lane_type == LaneType::SharedLeftTurn
+                    || lane.lane_type == LaneType::Bus
             })
             .count();
         tags.insert("lanes", lane_count.to_string());
@@ -662,15 +664,14 @@ pub fn lanes_to_tags(lanes: &[LaneSpec], _cfg: &Config) -> Result<Tags, LaneSpec
     }
     // Pedestrian
     {
-        match (lanes.first().unwrap().lane_type == LaneType::Sidewalk, lanes.last().unwrap().lane_type == LaneType::Sidewalk)
-        {
+        match (
+            lanes.first().unwrap().lane_type == LaneType::Sidewalk,
+            lanes.last().unwrap().lane_type == LaneType::Sidewalk,
+        ) {
             (false, false) => {}
-            (true, false) => assert!(tags.insert("sidewalk", "left")
-                .is_none()),
-            (false, true) => assert!(tags.insert("sidewalk", "right")
-                .is_none()),
-            (true, true) => assert!(tags.insert("sidewalk", "both")
-                .is_none()),
+            (true, false) => assert!(tags.insert("sidewalk", "left").is_none()),
+            (false, true) => assert!(tags.insert("sidewalk", "right").is_none()),
+            (true, true) => assert!(tags.insert("sidewalk", "both").is_none()),
         }
     }
     // Parking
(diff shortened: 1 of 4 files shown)


rust/osm2lanes/src/transform.rs
```diff
@@ -502,34 +512,129 @@ fn osm_separation_type(x: &str) -> Option<BufferType> {
...
     let mut tags = std::collections::BTreeMap::new();
+    let mut oneway = false;
     tags.insert("highway".to_owned(), "yes".to_owned()); // TODO, what?
+                                                         // Lanes
     {
         let lane_count = lanes
             .iter()
-            .filter(|lane| lane.lane_type == LaneType::Driving)
+            .filter(|lane| {
+                lane.lane_type == LaneType::Driving || lane.lane_type == LaneType::SharedLeftTurn
```
dabreegster, line 522: https://wiki.openstreetmap.org/wiki/Key:lanes#Description
Why not `LaneType::Bus`?

rust/osm2lanes/src/transform.rs
```diff
@@ -502,34 +512,129 @@ fn osm_separation_type(x: &str) -> Option<BufferType> {
...
     }
-    if lanes.first().unwrap().lane_type == LaneType::Sidewalk
-        && lanes.last().unwrap().lane_type == LaneType::Sidewalk
+    // Pedestrian
     {
-        tags.insert("sidewalk".to_owned(), "both".to_string());
+        match (lanes.first().unwrap().lane_type == LaneType::Sidewalk, lanes.last().unwrap().lane_type == LaneType::Sidewalk)
+        {
+            (false, false) => {}
+            (true, false) => assert!(tags.insert("sidewalk".to_owned(), "left".to_owned())
```
dabreegster, line 541: Might I suggest adding another helper to `Tags` along the lines of:
```
pub fn insert<K: Into<String>, V: Into<String>>(&mut self, k: K, v: V) {
        self.0.insert(k.into(), v.into());
}
```

And maybe it can also assert the key isn't present yet

rust/osm2lanes/src/transform.rs
```diff
@@ -502,34 +512,129 @@ fn osm_separation_type(x: &str) -> Option<BufferType> {
...
+                .is_none()),
+            (true, true) => assert!(tags.insert("sidewalk".to_owned(), "both".to_owned())
+                .is_none()),
+        }
+    }
+    // Parking
+    match (
+        lanes
+            .iter()
+            .take_while(|lane| lane.lane_type != LaneType::Driving)
```
dabreegster, line 553: This is a pretty easy-to-understand way of finding the parking lanes on both sides, I like it. In A/B Street, I tried to give every lane an orientation, including parking lanes, and there were lots of weird bugs -- like sometimes the parking lane not matching the orientation of the driving lane adjacent to it. (I think the bug was with [parking, one-way driving, parking])

But this side-steps all of that; the only thing that would get confusing is [d, p, p, d], but I've _almost_ never seen center parking like this, and OSM can't tag it anyway, so...

rust/osm2lanes/src/transform.rs
droogmic: TDD, I just didn't think about it

droogmic: - [ ] LaneType::Bus

rust/osm2lanes/src/transform.rs
droogmic: I need to refactor tags into trait implementations, will make this part of it :)

dabreegster:
I fixed up the merge conflicts here: https://github.com/a-b-street/osm2lanes/tree/rust_lanes_to_tags_v2
But didn't want to force push here and clobber everything. Does this merge look correct to you?

droogmic:
> I fixed up the merge conflicts here: https://github.com/a-b-street/osm2lanes/tree/rust_lanes_to_tags_v2 But didn't want to force push here and clobber everything. Does this merge look correct to you?

https://github.com/a-b-street/osm2lanes/commit/09b9be63cd9263180ec0260be3db07923ace5c40
That looks good :), clobber away

dabreegster:
Ah, the diff in `tests.json` is weird; fields reordered and typos reintroduced.. Also, some tests fail now. :(

I can take a look later night, or handover to you if you don't mind

dabreegster:
(But the web deployment looks better with this; the normalized tags are _way_ improved for the starter example)

PR #29, opened 2022-01-16
Add simple tests; fixes for `busway:*=lane`

enzet:
I want to add more simple test (that aren't based on real OSM ways). Hope this will simplify debugging.

### DIFF
--- a/data/tests.json
+++ b/data/tests.json
@@ -729,5 +729,31 @@
                 "direction": "forward"
             }
         ]
+    },
+    {
+        "skip_rust": true,
+        "description": "test for `busway:left=lane` and `busway:right=lane`",
+        "tags": {
+            "busway:right": "lane",
+            "busway:left": "lane",
+            "highway": "secondary",
+            "lanes": "3",
+            "oneway": "yes"
+        },
+        "driving_side": "right",
+        "output": [
+            {
+                "type": "bus_lane",
+                "direction": "backward"
+            },
+            {
+                "type": "travel_lane",
+                "direction": "forward"
+            },
+            {
+                "type": "bus_lane",
+                "direction": "forward"
+            }
+        ]
     }
 ]
--- a/data/tests.json
+++ b/data/tests.json
@@ -732,6 +732,27 @@
     },
     {
         "skip_rust": true,
+        "description": "test for `busway:left=lane`",
+        "tags": {
+            "busway:left": "lane",
+            "highway": "secondary",
+            "lanes": "2",
+            "oneway": "yes"
+        },
+        "driving_side": "right",
+        "output": [
+            {
+                "type": "bus_lane",
+                "direction": "backward"
+            },
+            {
+                "type": "travel_lane",
+                "direction": "forward"
+            }
+        ]
+    },
+    {
+        "skip_rust": true,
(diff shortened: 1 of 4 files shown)


data/tests.json
```diff
@@ -729,5 +729,52 @@
...
+            "busway:left": "lane",
+            "highway": "secondary",
+            "lanes": "2",
+            "oneway": "yes"
+        },
+        "driving_side": "right",
+        "output": [
+            {
+                "type": "bus_lane",
+                "direction": "backward"
```
dabreegster, line 746: Wait, why should this be backward? I thought that would be `busway:left=opposite_lane`

droogmic, line 746: Is this right?

If I compare this to the first example in this file, it means
`cycleway:left=lane` has an opposite direction than `busway:left=lane`

enzet: Seems like I'm wrong.

And what should it be without `oneway=yes`? (And presumably `lanes=4`.)

  - travel lane backward,
  - bus lane backward,
  - bus lane forward,
  - travel lane forward.

Am I right?

enzet: Please see my answer to @dabreegster.

dabreegster: Based on https://wiki.openstreetmap.org/wiki/Key:busway and https://wiki.openstreetmap.org/wiki/Key:lanes#Description, I think that:

- this example with `lanes=2, oneway=yes, busway:left=lane` should be [bus ↑, travel ↑]
- to flip the direction, it'd be `lanes=2, oneway=yes, busway:left=opposite_lane` should be [bus ↓, travel ↑]
- your question of `lanes=4, busway:left=lane` and right-handed driving would be... actually this feels a bit ambiguous. :) Is the bus lane on the inside or outside?  My best guess is [bus ↓, travel ↓, travel ↑, travel ↑].

droogmic: > `lanes=4, busway:left=lane`

Agreed with @dabreegster. I will also add a warning about the ambiguity...

enzet: Thank you! I think, now I get the idea. But still don't actually get the ambiguity. What is the alternative in the third example? Is it [travel ↓, bus ↓, travel ↑, travel ↑]?

droogmic: Ideally it would follow the same rules as cycleway?
https://wiki.openstreetmap.org/wiki/Key:cycleway:right:oneway

But it is ambiguous for me because no similar rules exist for busway that I can find.

dabreegster: Correct -- nothing in the OSM wiki seems to say whether the bus lane will be closer to the shoulder of the road or not. In most places, I think it'll be closer to the shoulder (usually for an easy connection to bus stops or a bus-only motorway entrance/exit), but I don't know how to get tag exceptions without using https://wiki.openstreetmap.org/wiki/Key:lanes:psv

droogmic: So I think I could see an argument for any of:
[travel ↓, bus ↓, travel ↑, travel ↑]
[bus ↕, travel ↓, travel ↑, travel ↑]
[travel ↓, bus ↕, travel ↑, travel ↑]

droogmic:
I'll let you re-open this when you are ready? It is going a bit stale...

PR #30, opened 2022-01-16
Rust: Refactor Tags

droogmic:
There are some generic type stuff in rust that I am not too happy with...

### DIFF
--- a/rust/osm2lanes/src/tags.rs
+++ b/rust/osm2lanes/src/tags.rs
@@ -1,0 +1,141 @@
+use serde::Deserialize;
+
+use std::collections::BTreeMap;
+
+/// A representation for a OSM tags key
+#[derive(Clone)]
+pub enum TagKey {
+    Static(&'static str),
+    String(String),
+}
+
+impl TagKey {
+    pub const fn from(string: &'static str) -> Self {
+        TagKey::Static(string)
+    }
+    pub fn as_str(&self) -> &str {
+        match self {
+            Self::Static(v) => v,
+            Self::String(v) => v.as_str(),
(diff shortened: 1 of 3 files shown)


rust/osm2lanes/src/tags.rs
```diff
@@ -0,0 +1,141 @@
...
+        let mut map = BTreeMap::new();
+        for line in s.lines() {
+            let (key, val) = line.split_once("=").ok_or("tag must be = separated")?;
+            map.insert(key.to_owned(), val.to_owned());
+        }
+        Ok(Self(map))
+    }
+}
+
+// TODO, shouldn't TagKey be passed by reference?
```
dabreegster, line 86: Yeah, that's what I'd expect. If you just switch from `Into` to [AsRef](https://doc.rust-lang.org/std/convert/trait.AsRef.html), that may work?

dabreegster (commented):
Hmm, let me see if I can take this refactor a bit further... ok if I push a commit onto this PR?

rust/osm2lanes/src/tags.rs
dabreegster: Hmm, not straightforward, not sure and have to get back to other work..

dabreegster (approved):
I'm happy with this if you are. Performance really is negligible, so I wouldn't worry about clones for now. As long as the API is easy to use, we can improve internals of Tags later

droogmic:
> Hmm, let me see if I can take this refactor a bit further... ok if I push a commit onto this PR?

I see the update, thanks! I forgot to update that part :P

PR #31, opened 2022-01-17
Rust: Web Site Roundtrip Support

droogmic:
As we said we wanted.

I haven't deployed an example yet, need to steel your action....

### DIFF
--- a/rust/osm2lanes-web/src/main.rs
+++ b/rust/osm2lanes-web/src/main.rs
@@ -8,7 +8,10 @@ use yew::prelude::*;
 
 mod draw;
 
-use osm2lanes::{get_lane_specs_ltr, Config, DrivingSide, LanePrintable, LaneSpec, Tags};
+use osm2lanes::{
+    get_lane_specs_ltr_with_warnings, lanes_to_tags, Config, DrivingSide, LanePrintable, LaneSpec,
+    Lanes, Tags,
+};
 
 // Use `wee_alloc` as the global allocator.
 #[global_allocator]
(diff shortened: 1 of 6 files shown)


rust/osm2lanes-web/src/main.rs
```diff
@@ -120,14 +175,64 @@ impl Component for App {
...
+        log::trace!("Calculate: {}", value);
+        match Tags::from_str(&value) {
+            Ok(tags) => match get_lane_specs_ltr_with_warnings(&tags, &CFG) {
+                Ok(lanes) => match lanes_to_tags(&lanes.lanes, &CFG) {
+                    Ok(tags) => Ok((lanes, tags)),
+                    Err(e) => Err(Ok((lanes, e.to_string()))),
+                },
+                Err(e) => Err(Err(e.to_string())),
+            },
+            Err(_) => Err(Err("parsing tags failed".to_owned())),
```
dabreegster, line 188: Wee nesting! Reasonable

dabreegster (approved):
> I haven't deployed an example yet, need to steel your action....

We could auto-push a deployment to a /dev or something for all PRs, but it might be overkill. It's minimal effort for me to clone the PR and build locally. Also technically a security risk -- anyone submitting a PR causes us to host a deployment with their code, which we haven't reviewed yet!

rust/osm2lanes-web/src/main.rs
droogmic: Yeah, Maybe Ill do the right thing in the future and ternary enum it, but it worked :P

PR #33, opened 2022-01-20
Rust: Bus and Bicycle Updates

droogmic:


rust/osm2lanes/src/tags.rs
```diff
@@ -105,11 +114,69 @@ impl ToString for Tags {
     }
 }
 
+#[derive(Clone, Default)]
+pub struct TagTree(BTreeMap<String, TagTreeVal>);
```
dabreegster, line 118: Some docs / examples? This would take something like `cycleway:left:separation:right` and recursively split it into pieces? So if you also had `cycleway:lift:separation:width`, you could traverse the tree and see `right` and `width` in the same place?

dabreegster (commented):
I think I likely never added tests for the cycleway separation stuff. Since there are no tests, the implementation is messy, and it doesn't handle left-handed traffic, I'm fine cleaning this up and starting it fresh

rust/osm2lanes/src/tags.rs
droogmic: correct. docs/tests do seem like a good idea :)

dabreegster (approved):
Splitting out the different PSV schemas is a good idea

PR #35, opened 2022-01-23
Rust: Proposal for Locale Builder

droogmic:


rust/osm2lanes/src/lib.rs
```diff
@@ -262,11 +239,8 @@ mod tests {
             if !test.skip.is_none() && test.skip.unwrap() {
                 continue;
             }
-            let cfg = Config {
-                driving_side: test.driving_side,
-                inferred_sidewalks: true,
-            };
-            let lanes = get_lane_specs_ltr(&test.tags, &cfg);
+            let lc = Locale::builder().driving_side(test.driving_side).build();
```
dabreegster, line 242: Style nit: I'd prefer spelling out `locale` or at least not using `lc` as an abbreviation. `lc` to some readers may feel too close to "lane-changing" :)

rust/osm2lanes/src/locale.rs
```diff
@@ -0,0 +1,78 @@
...
+
+impl Locale {
+    pub fn builder() -> Config {
+        Config::default()
+    }
+}
+
+/// Configuration to build locale, context about the place where an OSM way exists.
+#[derive(Default)]
+pub struct Config {
```
dabreegster, line 20: So defaults for lane widths, bikes allowed in bus lanes, typical colors for lane markings, etc can all live here. Sounds good to me.

dabreegster (approved):

PR #36, opened 2022-01-23
Rust: Refactor Errors and More

droogmic:


rust/osm2lanes/src/tags.rs
```diff
@@ -156,10 +165,23 @@ impl TagsRead for Tags {
...
     fn insert<K: Into<TagKey>, V: Into<String>>(&mut self, k: K, v: V) -> Option<String> {
         self.0.insert(k.into().as_str().to_owned(), v.into())
     }
+    fn checked_insert<K: Into<TagKey>, V: Into<String>>(
+        &mut self,
+        k: K,
+        v: V,
+    ) -> Result<String, TagError> {
+        self.insert(k, v)
+            .ok_or_else(|| TagError("duplicate key".to_owned()))
```
dabreegster, line 185: I was going to suggest a docstring, but then I got confused over the error case. If `self.insert(k, v).is_some()`, then it's an error. Is this logic backwards or am I misunderstanding the intention?

dabreegster (approved):
LGTM other than the docstring and/or behavior of `checked_insert`

rust/osm2lanes/src/tags.rs
droogmic: oh yeah, definitely, which reminds me that I also forgot to make use of this code :/

PR #39, opened 2022-01-24
Rust: Refactor Transform By Direction

droogmic:
I promise I'll start adding features soon, but refactoring is just so therapeutic...


dabreegster (approved):
Nice refactor! It's great to not just be pressured to do feature stuff all the time in a work setting, isn't it?

PR #43, opened 2022-01-26
Rust: Web: Add Driving Side Toggle

droogmic:


dabreegster (approved):
Nice!

I haven't looked to figure out why, or when the issue was introduced, but alignment seems off:
Screenshot from 2022-01-26 23-21-42

droogmic:
> Nice!
> 
> I haven't looked to figure out why, or when the issue was introduced, but alignment seems off: Screenshot from 2022-01-26 23-21-42

that was fixed in https://github.com/a-b-street/osm2lanes/pull/42
I don't know either when it happened, but couldn't be bothered to bisect...

PR #45, opened 2022-01-28
Rust: Refactor Tags To Lanes

droogmic:
Preparation for more business logic to be added.

I could start with busses next, but I think I would rather see what happens when I try to add some separators


droogmic:
I did however, royally screw up when picking which commit to refactor, so need to redo the work...

rust/osm2lanes-web/src/main.rs
```diff
@@ -8,10 +8,9 @@ use yew::prelude::*;
 
 mod draw;
 
-use osm2lanes::{
-    lanes_to_tags, tags_to_lanes_with_warnings, DrivingSide, Lane, LanePrintable, Lanes,
-    LanesToTagsConfig, Locale, Tags,
-};
+use osm2lanes::{lanes_to_tags, tags_to_lanes, LanesToTagsConfig, TagsToLanesConfig};
```
dabreegster, line 11: Curious what motivates this split? If the namespace feels like it's getting too big, maybe we shouldn't publish a flattened one, but have modules maybe?

rust/osm2lanes/src/lib.rs
```diff
@@ -119,7 +119,7 @@ impl LanePrintable for Lane {
                 designated: LaneDesignated::Bus,
                 ..
             } => '🚌',
-            Self::Shoulder => 'ˢ',
+            Self::Shoulder => '🛆',
```
dabreegster, line 122: Hmm, my browser and terminal both have trouble with this. Might just be an issue with my system

rust/osm2lanes/src/transform/tags_to_lanes/bicycle.rs
```diff
@@ -0,0 +1,125 @@
...
+
+pub fn bicycle(
+    tags: &Tags,
+    locale: &Locale,
+    oneway: bool,
+    forward_side: &mut Vec<Lane>,
+    backward_side: &mut Vec<Lane>,
+    warnings: &mut RoadWarnings,
+) -> ModeResult {
+    impl Tags {
```
dabreegster, line 11: Rust is crazy -- I didn't realize you could add a method to a struct just in the middle of some other method! Any reason to not put this outside of `fn bicycle` and just in this module scope?

dabreegster (approved):
Giant files being split always good :)

rust/osm2lanes-web/src/main.rs
droogmic: I'll be honest, I don't know how to do this properly in rust and have not yet looked it up.

Maybe I'll make it the topic of my next refactoring :)

rust/osm2lanes/src/lib.rs
dabreegster: I'm fine with this one, I'll see if I can install that font...

rust/osm2lanes/src/transform/tags_to_lanes/bicycle.rs
droogmic: When I implemented this, I assumed this falls under the general programming "smallest scope is better".

In this case though after having copied this function into its own file, I agree this impl should sit next to the function

PR #46, opened 2022-01-29
Rust: Add Separator Rendering

droogmic:
Work towards data driven rendering.

### DIFF
--- a/rust/osm2lanes-web/src/draw.rs
+++ b/rust/osm2lanes-web/src/draw.rs
@@ -3,129 +3,123 @@ use piet::{
     TextAttribute, TextLayoutBuilder,
 };
 
-use osm2lanes::{Lane, LaneDirection, LanePrintable};
+use osm2lanes::{Lane, LaneDirection, LanePrintable, MarkingColor, Metre, Road};
+
+// TODO: newtype + From?
+fn color_into(c: MarkingColor) -> Color {
+    match c {
+        MarkingColor::White => Color::WHITE,
+        MarkingColor::Yellow => Color::YELLOW,
+        MarkingColor::Red => Color::RED,
+    }
+}
+
+struct Scale(f64);
+
+impl Scale {
+    fn scale(&self, m: Metre) -> f64 {
+        self.0 * m.val()
+    }
+}
 
 pub fn lanes<R: RenderContext>(
     rc: &mut R,
     (canvas_width, canvas_height): (u32, u32),
-    lanes: &[Lane],
+    road: &Road,
 ) -> Result<(), Error> {
+    let canvas_width = canvas_width as f64;
+    let canvas_height = canvas_height as f64;
+    let default_lane_width = Lane::DEFAULT_WIDTH;
+
+    let grassy_verge = Metre::new(1.0);
+    let asphalt_buffer = Metre::new(0.5);
+
+    let scale = Scale(
+        canvas_width / (road.total_width() + 2.0 * grassy_verge + 2.0 * asphalt_buffer).val(),
+    );
+
+    // Background
     rc.clear(None, Color::OLIVE);
 
-    let _canvas_width = canvas_width as f64;
-    let canvas_height = canvas_height as f64;
-    let grassy_verge = 10.0;
-    let asphalt_buffer = 10.0;
-    let lane_width = 80.0;
     rc.fill(
         Rect::new(
-            grassy_verge,
+            scale.scale(grassy_verge),
             0.0,
-            (grassy_verge + asphalt_buffer) + (lanes.len() as f64 * lane_width) + asphalt_buffer,
+            scale.scale(grassy_verge + asphalt_buffer + road.total_width() + asphalt_buffer),
             canvas_height,
         ),
         &Color::BLACK,
     );
-    let x = grassy_verge + asphalt_buffer - 5.0;
-    rc.stroke(
-        Line::new(
-            Point { x, y: 0.0 },
-            Point {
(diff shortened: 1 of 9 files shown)


droogmic:
image

rust/osm2lanes-web/src/draw.rs
```diff
@@ -1,131 +1,148 @@
...
-                Point {
-                    x,
-                    y: 0.7 * canvas_height,
-                },
-                *direction,
-            )?;
+
+    let mut left_edge = grassy_verge + asphalt_buffer;
+
+    for lane in &road.lanes {
```
dabreegster, line 56: Cool, this left-to-right rendering is much closer to what A/B Street and Streetmix and such do! This shows the power of treating markings as a lane, very nice

rust/osm2lanes/src/lib.rs
```diff
@@ -23,6 +23,27 @@ pub struct Road {
...
+}
+
+impl Road {
+    pub fn total_width(&self) -> Metre {
+        self.lanes
+            .iter()
+            .map(|lane| match lane {
+                Lane::Separator { markings } => markings
+                    .iter()
+                    .map(|marking| marking.width.unwrap_or(Metre(0.2)))
```
dabreegster, line 39: Hmm you hardcoded this default in the web viewer too. How about `marking.get_width() -> Metre` that fills in the default?

rust/osm2lanes/src/lib.rs
```diff
@@ -143,6 +250,33 @@ impl LanePrintable for LaneDirection {
...
+impl MarkingStyle {
+    pub fn as_utf8(&self) -> char {
+        match self {
+            Self::SolidLine => '|',
+            Self::BrokenLine => '¦',
+            Self::DashedLine => ':',
+            Self::DottedLine => '᛫',
+            Self::KerbDown => '\\',
+            Self::KerbUp => '/',
+            _ => '|',
```
dabreegster, line 262: Tradeoff: wildcard matching makes adding a future case a little harder, because the compiler wont force you to think about all of the places the value is used. But that's not so bad for this debug text rendering

rust/osm2lanes/src/lib.rs
```diff
@@ -174,6 +308,46 @@ mod tests {
         rust: Option<RustTesting>,
     }
 
+    impl Road {
+        fn approx_eq(&self, other: &Self) -> bool {
```
dabreegster, line 312: How about a docstring: `/// Marking color only counts for equality when defined on both objects being compared`

rust/osm2lanes/src/transform/tags_to_lanes/separator.rs
```diff
@@ -0,0 +1,150 @@
...
+#[allow(clippy::needless_collect)]
+pub fn insert_separators(lanes: Lanes) -> LanesResult {
+    let Lanes { lanes, warnings } = lanes;
+    let left_edge = lane_to_edge_separator(lanes.first().unwrap());
+    let right_edge = lane_to_edge_separator(lanes.first().unwrap());
+    let separators: Vec<Option<Lane>> = lanes
+        .windows(2)
+        .map(|window| lanes_to_separator(window.try_into().unwrap(), &lanes))
+        .collect();
+    // I promise this is good code, but it might need a little explanation.
```
dabreegster, line 12: This is reasonable to understand. An alternate idea, take it or leave it: build up an output list as you go:

```
let mut output_lanes = Vec::new();
for pair in lanes.windows(2) {
  if output_lanes.is_empty() {
    output_lanes.push(lane_to_edge_separator(pair[0]));
  }
  output_lanes.push(pair[0]);
  if let Some(separator) = lanes_to_separator(pair) {
    output_lanes.push(separator);
  }  
}
output_lanes.push(lane_to_edge_separator(output_lanes.last().unwrap());
```

Maybe even simpler to do the left and right edge stuff outside the loop entirely

dabreegster (commented):
Looking awesome!

rust/osm2lanes/src/lib.rs
droogmic: indeed. also panics everyhwere that I need to remove.

rust/osm2lanes/src/lib.rs
droogmic: agreed

rust/osm2lanes/src/transform/tags_to_lanes/separator.rs
droogmic: not convinced by yours either. I am too proud of my iterator solution to delete it now :P

rust/osm2lanes/src/transform/tags_to_lanes/separator.rs
```diff
@@ -0,0 +1,150 @@
...
+        .chain(iter::once(right_edge))
+        .zip(lanes.into_iter().map(Some).chain(iter::once(None)))
+        .flat_map(|(a, b)| [a, b])
+        .flatten()
+        .collect();
+    Ok(Lanes { lanes, warnings })
+}
+
+/// Given a lane on the edge of a way
+/// what should the separator be
```
droogmic, line 26: wrong docstring

droogmic:
So, this is far from perfect, but I think I want to get this merged, and then do multiple rounds of refactoring now that I know what kinds of problems this is going to bring.

dabreegster (approved):
I'm always in favor of merging quickly and iterating :)  Awesome step forward!

PR #50, opened 2022-01-30
Rust: Refactor Modules

droogmic:
This contains no changes in functionality.


rust/osm2lanes-web/src/main.rs
```diff
@@ -9,9 +9,11 @@ use yew::prelude::*;
 
 mod draw;
 
-use osm2lanes::{lanes_to_tags, tags_to_lanes, LanesToTagsConfig, TagsToLanesConfig};
-use osm2lanes::{DrivingSide, Locale};
-use osm2lanes::{Lane, LanePrintable, Lanes, Road, Tags};
+use osm2lanes::road::{Lane, LanePrintable, Road};
+use osm2lanes::{
+    lanes_to_tags, tags_to_lanes, transform::Lanes, LanesToTagsConfig, TagsToLanesConfig,
```
dabreegster, line 14: nitpick: Any reason to split into separate lines for some of these? It could be one big `use osm2lanes::{DrivingSide, local, tags::Tags, road::{Lane, Road, LanePrintable}, ...}` or one line per submodule. Mixed is fine, just curious if there's motivation behind it.

(I really wish `cargo fmt` by default would be really strict about ordering imports, so this kind of thing could be ignored by humans)

dabreegster (approved):
Seems like a good directory structure. I never expected us to wind up with so many public types, but it's one of those subtly complex domains to model. :)

rust/osm2lanes-web/src/main.rs
droogmic: > (I really wish cargo fmt by default would be really strict about ordering imports, so this kind of thing could be ignored by humans)

Basically this, I don't know so I just do something that happens to look OK....

https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=import#imports_granularity

So let's just pick one?
I think I am partial to `Module` or `One`

droogmic: https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=import#group_imports

set this to StdExternalCrate as well

dabreegster: I'm in favor of both of these... but unfortunately, they seem to only be available for nightly. We could require stable to build/test, but ask developers to use nightly for `cargo fmt`. (And enforce the same in the `cargo fmt --check` github actions) Seem like a compromise?

> set this to StdExternalCrate as well

Oh wow, I've been doing this manually for months to match a collaborator who was using CLion...

> I think I am partial to Module or One

Slight preference for `Module`, but don't care much

droogmic: new PR for this.

PR #52, opened 2022-01-31
Rust: rustfmt toml

droogmic:


dabreegster:
LGTM! I have https://github.com/a-b-street/osm2lanes/commit/c6fd645987cdbd1d026fad92212036f8d59910fb which can be pushed after this is merged to update docs and the CI tests

droogmic:
> LGTM! I have [c6fd645](https://github.com/a-b-street/osm2lanes/commit/c6fd645987cdbd1d026fad92212036f8d59910fb) which can be pushed after this is merged to update docs and the CI tests

missing an update to the git hook? that is what I use to qualify my pushes :)

(Side note: I dislike actions because a workflow cannot be run locally as a git hook, so we have 2 files to do the same thing)

PR #53, opened 2022-01-31
Use nightly for cargo fmt check in CI

dabreegster:
I wasn't familiar with git hooks, but they indeed have advantages. @enzet, any thoughts about consolidating the github actions and git hooks?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,6 +37,7 @@ jobs:
       uses: hecrj/setup-rust-action@v1
       with:
         rust-version: nightly
+        components: rustfmt
     - name: Test the Rust implementation
       run: |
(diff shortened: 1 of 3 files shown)


dabreegster:
Urgh, the setup rust action is harder to understand than `rustup`. It's switching the default to nightly. I'll work on this later

enzet:
I'm not familiar with cargo, but the the idea of keeping them consistent with GitHub Actions looks right.

The only problem with Git hooks is that they have be enabled by committer manually with

```shell
git config --local core.hooksPath data/githooks
```

I think we should add it to `README.md` or to `CONTRIBUTING.md`.

droogmic:
Fixed by #58

PR #54, opened 2022-01-31
Rust: Restructure and add Stubs for Tags To Lanes

droogmic:


rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -255,3 +183,91 @@ fn assemble_ltr(
...
+            RoadMsg::Unimplemented {
+                description: Some("lanes=*".to_owned()),
+                // TODO, TagTree should support subset
+                tags: Some(tags.subset(&["lanes"])),
+            }
+            .into(),
+        );
+    }
+
+    const ACCESS_KEYS: [&'static str; 43] = [
```
dabreegster, line 208: Could be helpful to document a source for this kind of info, like https://taginfo.openstreetmap.org/keys/access#values

dabreegster (approved):

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
droogmic, line 208:
```suggestion
    // https://wiki.openstreetmap.org/wiki/Key:access#Transport_mode_restrictions
    const ACCESS_KEYS: [&'static str; 43] = [
```

PR #55, opened 2022-01-31
Rust: Add Overpass and Async

droogmic:
move main.rs into its own crate...

2 subcommands, `convert` and `get` or something

and try to add a input form to the web view to also query overpass, just to confirm the reqwest library works in WASM.

probably not included: driving_side and iso3166 retrieval from overpass


rust/osm2lanes/Cargo.toml
```diff
@@ -9,3 +9,7 @@ authors = ["Dustin Carlino <dabreegster@gmail.com>"]
 log = "0.4"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
+reqwest = { version = "0.11", features = ["blocking", "json"] }
+
+[features]
+overpass = []
```
droogmic, line 15: connect these? but how?

dabreegster: Ah, you can do something like

```
request = { version = "0.11", optional = true }

[features]
overpass = ["reqwest", "reqwest/blocking", "reqwest/json"]
```

Just guessing that's what you're looking for?

droogmic: yup thanks

droogmic:
So, in this PR ill move main.rs into its own crate...

2 subcommands `convert` and `get` or something

and try to add a input form to the web view to also query overpass, just to confirm the reqwest library works in WASM.

probably late deliverable: driving_side and iso3166 retrieval

dabreegster:
For the CLI: sounds like we could either have two binaries, or one using clap/structopt/whatever else to have two different subcommands. If you wanted to go with the former and avoid another crate, you can have multiple binaries in a crate: https://doc.rust-lang.org/cargo/guide/project-layout.html

For web: reqwest should work in wasm. Example at https://github.com/a-b-street/abstreet/blob/e5704f4a6d2e742924ed43be87d54f5a4b354ee5/abstio/src/http.rs#L23 -- figuring out all the combinations of IO and networking in native/web definitely can be a pain. IIRC, `blocking` in reqwest may not work on wasm, because the browser runtime doesn't really let you block. Does yew have async support?

droogmic:
> For the CLI: sounds like we could either have two binaries, or one using clap/structopt/whatever else to have two different subcommands. If you wanted to go with the former and avoid another crate, you can have multiple binaries in a crate: https://doc.rust-lang.org/cargo/guide/project-layout.html

I didn't know bin/ was a thing. I guess that is better?
Does it make osm2lanes as a rust dependency bigger, or is it hidden from the API?

> For web: reqwest should work in wasm. Example at https://github.com/a-b-street/abstreet/blob/e5704f4a6d2e742924ed43be87d54f5a4b354ee5/abstio/src/http.rs#L23 -- figuring out all the combinations of IO and networking in native/web definitely can be a pain. IIRC, `blocking` in reqwest may not work on wasm, because the browser runtime doesn't really let you block. Does yew have async support?

Async yew and reqwest are all things I need to test, I don't know :P

dabreegster:
> Does it make osm2lanes as a rust dependency bigger, or is it hidden from the API?

Unfortunately yes. You can kind of hack around it by making the bin-only dependencies optional, but when you do `cargo run --bin foo`, it'll just not work without passing `--features=overpass` or whatever. There's some bug to add support for doing this more easily in Cargo. But... a separate crate makes sense I guess. It's possible some people will actually want to consume osm2lanes as a CLI tool; that could be easier than Python FFI bindings or something. So having a more separated space to make a nice CLI would be useful.

droogmic:
Looking good now, marking as ready for review, although I still need to do a self review.

rust/osm2lanes-bin/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    #[clap(setting(AppSettings::ArgRequiredElseHelp))]
+    Convert {
+        /// JSON of OSM Tags
+        #[clap(required = true, parse(from_os_str))]
+        path: PathBuf,
+    },
+}
+
+#[tokio::main]
+async fn main() {
```
droogmic, line 40: I tried to look for a lightweight non-spinning async executor that only supports `block_on`, but I had no luck....

rust/osm2lanes-web/src/main.rs
```diff
@@ -115,11 +116,30 @@ impl Component for App {
...
+                    .state
+                    .way_ref
+                    .cast::<HtmlInputElement>()
+                    .unwrap()
+                    .value();
+                log::debug!("WayFetch {}", way_id);
+                match way_id.parse() {
+                    Ok(way_id) => {
+                        ctx.link().send_future(async move {
+                            let tags = get_way(way_id).await.unwrap();
```
droogmic, line 130: - [ ] handle error, remove unwrap

rust/osm2lanes-bin/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+use std::io::BufReader;
+use std::path::PathBuf;
+
+use clap::{AppSettings, Parser, Subcommand};
+use futures::executor::block_on;
+use osm2lanes::overpass::get_way;
+use osm2lanes::tags::Tags;
+use osm2lanes::{tags_to_lanes, Locale, TagsToLanesConfig};
+
+/// Simple program to greet a person
```
dabreegster, line 11:
```suggestion
/// Generate lanes from an OSM way
```

rust/osm2lanes-bin/src/main.rs
dabreegster: It sort of feels like tokio is emerging as the standard now. I think with the right feature flags, it shouldn't add too much compile time or final binary size

rust/osm2lanes/src/overpass.rs
```diff
@@ -0,0 +1,43 @@
...
+#[derive(Debug, Clone, Deserialize)]
+struct OverpassResult {
+    // version: f32,
+    // osm3s: Osm3s
+    elements: Vec<Element>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+struct Element {
+    r#type: ElementType,
```
dabreegster, line 14: Whoaaaa is that how you use reserved words as field names? I never knew! (I've always done `#[serde(rename = "type")]` and named the field something else

dabreegster (approved):

PR #59, opened 2022-02-03
Rust: Refactor Tranform Error

droogmic:
The important functional changes:

- How we serialize errors (as display)

### DIFF
--- a/rust/osm2lanes/src/transform/error.rs
+++ b/rust/osm2lanes/src/transform/error.rs
@@ -35,10 +35,7 @@ pub enum RoadMsg {
         tags: Option<Tags>,
     },
     /// Other issue
-    Other {
-        description: String,
-        tags: Tags,
-    },
+    Other { description: String, tags: Tags },
     /// Internal error
     TagsDuplicateKey(DuplicateKeyError),
 }
--- a/rust/osm2lanes/src/transform/error.rs
+++ b/rust/osm2lanes/src/transform/error.rs
@@ -14,32 +14,32 @@ use crate::tags::{DuplicateKeyError, TagKey, Tags};
 ///
 #[derive(Debug, Clone)]
 pub enum RoadMsg {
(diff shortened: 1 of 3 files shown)


rust/osm2lanes/src/transform/error.rs
```diff
@@ -0,0 +1,198 @@
...
+/// ```
+/// use osm2lanes::transform::RoadMsg;
+/// let _ = RoadMsg::deprecated_tag("foo", "bar");
+/// let _ = RoadMsg::unsupported_tag("foo", "bar");
+/// let _ = RoadMsg::unsupported_str("foo=bar because x and y");
+/// ```
+///
+#[derive(Debug, Clone)]
+pub enum RoadMsg {
+    // Deprecated OSM tags, with suggested alternative
```
dabreegster, line 17: nit: Consider `///` for these, they'd make good docstrings

rust/osm2lanes/src/transform/error.rs
```diff
@@ -0,0 +1,198 @@
...
+                .map(|warning| format!("Warning: {}", warning))
+                .collect::<Vec<_>>()
+                .join("\n")
+        )
+    }
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub enum RoadError {
+    #[serde(serialize_with = "use_display")]
```
dabreegster, line 158: Is this redundant? There's an explicit `impl Serialize for RoadMsg`.

... I got curious, and it does seem redundant. I'd suggest adding a doc test above to show the JSON serialization:

```
/// let err = RoadError::from(RoadMsg::deprecated_tag("foo", "bar"));
/// assert_eq!("{\"Msg\":\"unsupported: foo=bar\"}", serde_json::to_string(&err).unwrap());
```

dabreegster (approved):

dabreegster (approved):

PR #60, opened 2022-02-05
Rust NPM Package

droogmic:
In response to https://github.com/3DStreet/3dstreet/issues/138#issuecomment-1030523013


dabreegster:
I wonder if following this template is overkill. New dependencies on Travis and Appveyor build systems? We don't need full automation for publishing to NPM every time we commit here; we don't _want_ that even! To start, if we manually run a few commands to publish, that'd almost be preferable.

Also not sure we need a new crate to expose a JS interface. In another project, I just tacked on the tiny WASM interface at the end: https://github.com/zonebuilders/zonebuilder-rust/blob/a3070b9332227f9ebc074c6879448d050ea99887/src/lib.rs#L272

I may have some time this weekend to play around with all of this. I'm fine moving forward with something working, but I would like to eventually go back and simplify it

droogmic:
> I wonder if following this template is overkill. New dependencies on Travis and Appveyor build systems? We don't need full automation for publishing to NPM every time we commit here; we don't _want_ that even! To start, if we manually run a few commands to publish, that'd almost be preferable.

Oh, I need to cleanup the template (draft). I wuold expect a simple Github Actions based deployment...

> Also not sure we need a new crate to expose a JS interface. In another project, I just tacked on the tiny WASM interface at the end: https://github.com/zonebuilders/zonebuilder-rust/blob/a3070b9332227f9ebc074c6879448d050ea99887/src/lib.rs#L272

Agreed, but I don't want to bring any WASM stuff into the default osm2lanes crate.
So either tack it onto osm2lanes-web or make it a feature flag on osm2lanes? I like the sound of the second option, if we can do it in a way that is hidden from rust x86 clients.

dabreegster:
> So either tack it onto osm2lanes-web or make it a feature flag on osm2lanes? 

`osm2lanes-web` is all about rendering / Yew; it's an "end-user app" at the moment, so probably not there.

The feature flags should be pretty simple: https://github.com/zonebuilders/zonebuilder-rust/blob/a3070b9332227f9ebc074c6879448d050ea99887/Cargo.toml#L21

So ideally all we need to do is add that to `osm2lanes` crate, do `wasm-pack build --release`, then do a few things to take the generated directory and get it on NPM. I don't know how NPM works yet, so maybe it's more elaborate than that

rust/osm2lanes/Cargo.toml
```diff
@@ -1,18 +1,30 @@
...
+console_error_panic_hook = { version = "0.1", optional = true }
 log = "0.4"
-serde = { version = "1", features = ["derive"] }
 reqwest = { version = "0.11", optional = true, features = ["blocking", "json"] }
+serde = { version = "1", features = ["derive"] }
+wasm-bindgen = { version = "0.2", optional = true, features = ["serde-serialize"] }
+wee_alloc = { version = "0.4", optional = true }
 
 [features]
+default = ["console_error_panic_hook"]
```
dabreegster, line 23: Wait, why's this in default? Related to trying to make features per compilation target work?

dabreegster (commented):
This seems like what I'd expect to need. I should just run things myself and see, but I don't understand what the problem is with tying everything to `--features=wasm`. But a separate crate also is fine if that's needed

PR #61, opened 2022-02-05
Rust: Refactor Road Markings

droogmic:


rust/osm2lanes/src/road/marking.rs
```diff
@@ -1,7 +1,39 @@
...
+
+#[derive(Clone, PartialEq, Debug, Serialize, Deserialize)]
+pub struct Markings(Vec<Marking>);
+
+impl Markings {
+    pub fn new(markings: Vec<Marking>) -> Self {
+        Self(markings)
+    }
+
+    /// Reverses the order of markings and inverts them in place.
```
dabreegster, line 14: Example of when we'd use this? I was picturing `invert` as changing the direction of one-way arrows or something, but it just swaps up/down kerbs. But I think I don't understand the bigger picture of when we'd want to reverse Markings

dabreegster (approved):

rust/osm2lanes/src/road/marking.rs
droogmic: Hmm, maybe I gave up on the branch that needed this and never pushed it, but what I found was the following:

1. lanes: forwards/backwards, out to middle
2. pair of lanes: out to middle -> separator (so we only have to handle one side of the road
2. lanes with separators: forwards/backwards, out to in
3. construct lanes like https://github.com/a-b-street/osm2lanes/blob/main/rust/osm2lanes/src/transform/tags_to_lanes/mod.rs#L168

problem: inverting separators is not a trivial operation like the other lanes, they are handed/chiral.

PR #66, opened 2022-02-07
Rust Website: Explicitly Enumerate Warnings

droogmic:


rust/osm2lanes-web/src/main.rs
```diff
@@ -279,29 +279,28 @@ impl Component for App {
...
+        log::trace!("Update Tags: {}", value);
+        let calculate = match Tags::from_str(value) {
             Ok(tags) => match tags_to_lanes(&tags, locale, &TagsToLanesConfig::default()) {
                 Ok(lanes) => {
                     match lanes_to_tags(&lanes.lanes, locale, &LanesToTagsConfig::default()) {
                         Ok(tags) => Ok((lanes, tags)),
-                        Err(e) => Err(Ok((lanes, e.to_string()))),
+                        Err(e) => {
+                            if let RoadError::Warnings(warnings) = &e {
+                                Err(Ok((lanes, format!("{}\n{}", e, warnings))))
```
dabreegster, line 293: Would this wind up repeating something? If you've got a before/after screenshot or example, that'd be helpful, or I can run locally and see...

droogmic: it does often wind-up repeating warnings, but this is important for debugging, because they do not have to be the same.

Warning generated in the first tags(input) to lanes,
are not the same as the warnings generated after tags(input) -> lanes -> tags -here-> lanes (rountrip testing the lanes to tags step).

dabreegster (approved):
LGTM

PR #70, opened 2022-02-07
Rework Git Hooks and Actions

droogmic:
Split stuff up
Add caching to python
Make scripts the same between hooks and actions.

I still dislike actions.

### DIFF
--- a/.githooks/check
+++ b/.githooks/check
@@ -1,29 +1,26 @@
 #!/bin/sh
 
-# Spec.
+set -e -u -o pipefail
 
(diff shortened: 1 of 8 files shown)


dabreegster (approved):
IMO: Actions can be convenient for deployment and CI, but the main convenience comes from the integration with github and being free. The fact that there's a bespoke config language, and you have to reason about the runners as their own environment with certain software versions of stuff, is unfortunate. I would almost prefer they use something more standard, like just running some Dockerfiles.

Thanks for refactoring to support them along with git hooks! A good investment for other people to make contributions if they forget to set up git hooks locally

PR #73, opened 2022-02-08
Rust: Tags Refactor

droogmic:


droogmic:
I had to redo the CI again... still not happy with it :)

droogmic:
> I had to redo the CI again... still not happy with it :)

(it was not reporting errors...)

.githooks/check
```diff
@@ -1,26 +1,38 @@
-#!/bin/sh
+#!/bin/env bash
```
dabreegster, line 1: I feel like we're reinventing test runners :(  But I don't know a simpler approach. Shell not propagating errors nicely despite `set -e` is definitely frustrating

dabreegster (approved):
If it's not loads of trouble next time, splitting up refactoring and CI fixes would be preferable. Fine here, but if it's ever a big change / heavy feature, a focused diff can be useful.

Thanks for all of the work so far!

droogmic:
> If it's not loads of trouble next time, splitting up refactoring and CI fixes would be preferable. Fine here, but if it's ever a big change / heavy feature, a focused diff can be useful.
> 
> Thanks for all of the work so far!

agreed, laziness strikes again

.githooks/check
droogmic: tbh, I am not liking my refactoring.

I might revert the effort I put into making the githooks and actions share the same source of truth. I just don't think it is worth the effort :(

PR #75, opened 2022-02-09
Rust: Refactor Tags, Part 1?

droogmic:
Make TagKey enum private, to allow for future changes in implementation
Remove TagRead trait, as unecessary.
Remove public interfaces that used BTreeMap.

### DIFF
--- a/rust/osm2lanes/src/tag/mod.rs
+++ b/rust/osm2lanes/src/tag/mod.rs
@@ -24,6 +24,7 @@ impl std::fmt::Display for DuplicateKeyError {
 pub struct Tags(BTreeMap<String, String>);
 
 impl Tags {
+    /// Construct from slice of pairs
     pub fn from_str_pairs(tags: &[[&str; 2]]) -> Result<Self, DuplicateKeyError> {
         let mut map = BTreeMap::new();
         for tag in tags {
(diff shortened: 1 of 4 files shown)


rust/osm2lanes/src/tag/mod.rs
```diff
@@ -33,16 +34,12 @@ impl Tags {
...
-    // return something like &[[&str; 2]]
-    /// Expose inner map
-    pub fn map(&self) -> &BTreeMap<String, String> {
-        &self.0
+    /// Expose data as vector of pairs
+    pub fn to_str_pairs(&self) -> Vec<[&str; 2]> {
+        self.0
+            .iter()
+            .map(|(k, v)| [k.as_str(), v.as_str()])
+            .collect::<Vec<[&str; 2]>>()
```
dabreegster, line 42: Type inference may let you do `collect::<Vec<_>>()` here and avoid repeating the return value

rust/osm2lanes/src/tag/mod.rs
```diff
@@ -53,6 +50,50 @@ impl Tags {
             .collect::<Vec<String>>()
     }
 
+    // TODO, shouldn't TagKey be passed by reference?
```
dabreegster, line 53: I feel like I tried to do this conversion a few weeks back and failed, but theoretically `AsRef<TagKey>` could work

dabreegster (approved):

rust/osm2lanes/src/tag/mod.rs
droogmic: I managed: https://github.com/droogmic/osm2lanes/pull/1/files

rust/osm2lanes/src/tag/mod.rs
droogmic:
```suggestion
            .collect::<Vec<_>>()
```

PR #77, opened 2022-02-10
Rust: Lane Width

droogmic:
Add lane width to rust lane structs.
Locale now does the heavy lifting with travel_width.

### DIFF
--- a/rust/osm2lanes/src/locale.rs
+++ b/rust/osm2lanes/src/locale.rs
@@ -16,7 +16,7 @@ impl Locale {
     pub fn builder() -> Config {
         Config::default()
     }
-    pub fn default_width(&self, designated: &LaneDesignated) -> Metre {
+    pub const fn travel_width(&self, designated: &LaneDesignated) -> Metre {
         match designated {
             LaneDesignated::Motor | LaneDesignated::Bus => Metre::new(3.5),
             LaneDesignated::Foot => Metre::new(2.5),
--- a/rust/osm2lanes/src/locale.rs
+++ b/rust/osm2lanes/src/locale.rs
@@ -20,7 +20,7 @@ impl Locale {
     pub fn builder() -> Config {
         Config::default()
     }
-    pub const fn travel_width(&self, designated: &LaneDesignated) -> Metre {
+    pub fn travel_width(&self, designated: &LaneDesignated) -> Metre {
         match designated {
             LaneDesignated::Motor | LaneDesignated::Bus => Metre::new(3.5),
             LaneDesignated::Foot => Metre::new(2.5),
--- a/rust/osm2lanes/src/road/lane.rs
+++ b/rust/osm2lanes/src/road/lane.rs
@@ -12,14 +12,16 @@ pub enum Lane {
         // TODO, we could make this non-optional, but remove the field for designated=foot?
         direction: Option<LaneDirection>,
         designated: LaneDesignated,
+        width: Option<Metre>,
     },
     #[serde(rename = "parking")]
     Parking {
         direction: LaneDirection,
         designated: LaneDesignated,
+        width: Option<Metre>,
     },
     #[serde(rename = "shoulder")]
-    Shoulder,
+    Shoulder { width: Option<Metre> },
     #[serde(rename = "separator")]
     Separator { markings: Markings },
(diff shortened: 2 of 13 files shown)


rust/osm2lanes/src/locale.rs
```diff
@@ -16,7 +16,7 @@ impl Locale {
     pub fn builder() -> Config {
         Config::default()
     }
-    pub fn default_width(&self, designated: &LaneDesignated) -> Metre {
+    pub const fn travel_width(&self, designated: &LaneDesignated) -> Metre {
```
dabreegster, line 19: Isn't it likely we'll break `const` in the near future? If locale starts doing any logic to use config for different places, we might need something simple, even a `HashMap`, that isn't const yet

rust/osm2lanes/src/road/lane.rs
```diff
@@ -33,8 +35,14 @@ impl Lane {
     pub fn width(&self, locale: &Locale) -> Metre {
         match self {
             Lane::Separator { markings } => markings.width(locale),
-            Lane::Travel { designated, .. } => locale.default_width(designated),
-            _ => Lane::DEFAULT_WIDTH,
+            Lane::Travel {
+                width, designated, ..
+            } => width.unwrap_or_else(|| locale.travel_width(designated)),
+            // TODO: parking different from travel?
```
dabreegster, line 41: 100% yes, parking lanes are way more narrow than travel lanes in the UK a tleast

rust/osm2lanes/src/road/lane.rs
```diff
@@ -12,14 +12,16 @@ pub enum Lane {
         // TODO, we could make this non-optional, but remove the field for designated=foot?
         direction: Option<LaneDirection>,
         designated: LaneDesignated,
+        width: Option<Metre>,
```
dabreegster, line 15: So what's the overall picture -- if we can scrape lane width from OSM, this is filled out, but otherwise the caller can use `Locale::travel_width`?

A thing to think about for the near future: in OSM, usually width of the entire road is mapped (and of course it's a bit complicated what "entire road" actually covers). We may need a way to stash that, and to think about algorithms that can divide the total width across lanes in some reasonable way (probably by guessing a default width by lane type, adding that up, taking a ratio of that sum to the total in OSM, and scaling appropriately)

rust/osm2lanes/src/locale.rs
droogmic:
```suggestion
    pub fn travel_width(&self, designated: &LaneDesignated) -> Metre {
```

rust/osm2lanes/src/transform/mod.rs
```diff
@@ -71,34 +71,47 @@ impl Lane {
...
             designated,
+            width: Some(locale.travel_width(&designated)),
         }
     }
-    fn both(designated: LaneDesignated) -> Self {
+    fn both(designated: LaneDesignated, locale: &Locale) -> Self {
         Self::Travel {
             direction: Some(LaneDirection::Both),
             designated,
+            width: Some(locale.travel_width(&designated)),
```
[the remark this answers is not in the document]
droogmic: our tests still have width unspecified.

Internally, I think this should never be None...

I think at some point I will need to separate the serde struct from the internal type, for reasons like this.
Downside is an explosion of glue code...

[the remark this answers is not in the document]
droogmic: I made https://github.com/a-b-street/osm2lanes/issues/79

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -116,25 +116,44 @@ enum LaneType {
     Shoulder,
 }
 
+#[derive(Default)]
+struct Width {
+    min: Infer<Metre>,
+    target: Infer<Metre>,
```
dabreegster, line 122: Just so I understand, `target` is our best guess, but we can specify bounds?

dabreegster (approved):

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
droogmic: correct

PR #81, opened 2022-02-11
Rust Website: Add Country Support

droogmic:
Uses celec dependency for ISO 3166 country definitions.

The default selected country should maybe be set initially, but for now is not.

### DIFF
--- a/rust/osm2lanes-web/src/main.rs
+++ b/rust/osm2lanes-web/src/main.rs
@@ -3,18 +3,17 @@ use std::str::FromStr;
 use piet::Error as PietError;
 use piet_web::WebRenderContext;
 use wasm_bindgen::JsCast;
-use web_sys::{window, HtmlCanvasElement, HtmlInputElement};
+use web_sys::{window, HtmlCanvasElement, HtmlInputElement, HtmlSelectElement};
 use yew::prelude::*;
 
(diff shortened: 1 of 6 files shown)


rust/osm2lanes/src/lib.rs
```diff
@@ -11,13 +11,13 @@ pub mod road;
 
 pub mod tag;
 
-mod locale;
-pub use self::locale::{DrivingSide, Locale};
+pub mod locale;
+pub use locale::{DrivingSide, Locale};
```
droogmic, line 15: I will later deprecate these crate-root exports

rust/osm2lanes-web/src/main.rs
```diff
@@ -109,6 +109,15 @@ impl Component for App {
...
             }
+            Msg::CountrySet(Ok(country)) => {
+                self.state.locale.country = Some(country);
+                self.update_tags();
+                true
+            }
+            Msg::CountrySet(Err(country_err)) => {
+                self.state.message = Some(country_err.to_owned());
+                true
+            }
```
droogmic, line 120: maybe unnecessary, would only be triggered by client manipulation of the countries...

dabreegster (approved):
Nice and simple :)

PR #84, opened 2022-02-16
Add Support for More Bus Lanes

droogmic:
Expand `busway:*=opposite_lane` and rework `bus:lanes=*`

### DIFF
--- a/rust/osm2lanes/src/transform/lanes_to_tags.rs
+++ b/rust/osm2lanes/src/transform/lanes_to_tags.rs
@@ -168,11 +168,33 @@ pub fn lanes_to_tags(lanes: &[Lane], locale: &Locale, config: &LanesToTagsConfig
             .rev()
             .take_while(|lane| !lane.is_motor())
             .find(|lane| lane.is_bus());
-        match (left_bus_lane.is_some(), right_bus_lane.is_some()) {
-            (false, false) => {}
-            (true, false) => tags.checked_insert("busway:left", "lane")?,
-            (false, true) => tags.checked_insert("busway:right", "lane")?,
-            (true, true) => tags.checked_insert("busway:both", "lane")?,
+        if left_bus_lane.is_none()
+            && right_bus_lane.is_none()
+            && lanes.iter().any(|lane| lane.is_bus())
+        {
+            tags.checked_insert(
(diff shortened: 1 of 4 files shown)


rust/osm2lanes/src/transform/lanes_to_tags.rs
```diff
@@ -174,11 +174,33 @@ pub fn lanes_to_tags(lanes: &[Lane], locale: &Locale, config: &LanesToTagsConfig
...
+        if left_bus_lane.is_none()
+            && right_bus_lane.is_none()
+            && lanes.iter().any(|lane| lane.is_bus())
+        {
+            tags.checked_insert(
+                "bus:lanes",
+                lanes
+                    .iter()
+                    .map(|lane| if lane.is_bus() { "designated" } else { "" })
+                    .collect::<Vec<_>>()
```
dabreegster, line 186: I don't know a better way to do this, just lamenting that it's not easier to build up the string without first making a full vector (and then doing `as_slice`). Might all be possible, but the error messages aren't simple if so. I miss Scala sometimes for reasons like this :\

rust/osm2lanes/src/transform/tags_to_lanes/bus.rs
```diff
@@ -145,56 +171,149 @@ fn lanes_bus(
...
+                    match access {
+                        Access::None => {}
+                        Access::No => {}
+                        Access::Yes => {}
+                        Access::Designated => lane.set_bus(locale)?,
+                    }
                 }
             }
         }
+        // Don't try to understand this
```
dabreegster, line 297: I had no intention to; I trust the tests ;)

dabreegster (approved):
Awesome! I am increasingly motivated to cutover A/B Street and take advantage of all the progress here!

rust/osm2lanes/src/transform/lanes_to_tags.rs
droogmic: I always just assumed the implementation required a sequence in memory, but you made me go luck, and it doesn't appear to be the case: https://doc.rust-lang.org/src/alloc/slice.rs.html#776

Maybe my first PR to rust can be to try and add join to the Iterator trait...

PR #86, opened 2022-02-21
Add Highway Support

droogmic:
A follow-up will be needed to structurally solve roundtrip information retention

### DIFF
--- a/data/tests.yml
+++ b/data/tests.yml
@@ -1113,19 +1113,19 @@
 - way_id: 323605308
   rust: false
   tags:
-    highway: primary
-    bus:lanes: yes|yes|designated
-    busway:right: lane
-    cycleway:right: share_busway
+    highway: 'primary'
+    bus:lanes: 'yes|yes|designated'
+    busway:right: 'lane'
+    cycleway:right: 'share_busway'
     lanes: '3'
     lanes:psv: '1'
     lit: 'yes'
     maxspeed: '30'
-    name: Quai Aimé Césaire
+    name: 'Quai Aimé Césaire'
     oneway: 'yes'
-    psv:lanes: yes|yes|designated
-    surface: asphalt
-    taxi:lanes: yes|yes|designated
+    psv:lanes: 'yes|yes|designated'
+    surface: 'asphalt'
+    taxi:lanes: 'yes|yes|designated'
   driving_side: right
   output:
   - type: travel
--- a/data/tests.yml
+++ b/data/tests.yml
@@ -210,7 +210,7 @@
   skip_python: true
   skip_kotlin: true
   tags:
-    highway: 'yes'
+    highway: 'road'
     lanes: '1'
     oneway: 'yes'
     shoulder: 'no'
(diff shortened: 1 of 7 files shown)


data/tests.yml
```diff
@@ -1107,10 +1113,10 @@
 - way_id: 323605308
   rust: false
   tags:
+    highway: primary
```
dabreegster, line 1116: No quotes here?

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -363,14 +363,31 @@ fn assemble_ltr(
...
+                | Highway::Residential
+                // Non motorized
+                | Highway::Pedestrian
+                | Highway::Footway
+                | Highway::Cycleway
+                | Highway::Path
+                | Highway::Steps
+                | Highway::Track
+             => highway,
+            highway => return Err(RoadMsg::unimplemented_tag(HIGHWAY, &highway.to_string()).into()),
```
dabreegster, line 388: What's happening here?

dabreegster (commented):
I think `osm.rs` is missing? I was trying to comment out one of the match lines, but I can't even build

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
droogmic: if it is not a known supported enum variant, we say it is not implemented.
This replaces the previous construction and guided busway placeholders

droogmic:
Oops, there you go, sorry :(

rust/osm2lanes/src/tag/osm.rs
```diff
@@ -0,0 +1,136 @@
...
+    Corridor,
+    Cycleway,
+    Escape,
+    Footway,
+    LivingStreet,
+    Path,
+    Pedestrian,
+    Proposed,
+    Raceway,
+    Residential,
```
dabreegster, line 20: Why isn't residential considered a road? Or service roads? (What's the difference between Road and Highway -- is it just stuff that can have a `_link` suffix?)

rust/osm2lanes/src/tag/osm.rs
```diff
@@ -0,0 +1,136 @@
...
+            "residential" => Self::Residential,
+            "service" => Self::Service,
+            "steps" => Self::Steps,
+            "track" => Self::Track,
+            _ => return Err(s.to_owned()),
+        })
+    }
+}
+
+impl std::fmt::Display for Highway {
```
dabreegster, line 88: It would be great if we could specify these when we declare the enum and have a from/to-string automatically. I've used https://docs.rs/strum_macros/0.23.1/strum_macros/derive.EnumString.html before, but not entirely sure this lets you do both directions

dabreegster: Fine by me, just a thought. Maybe one day I'll play with proc macros finally and make something like this, if strum indeed can't

dabreegster:
LGTM! Big step forwards to not saying a building has 2 lanes ;)

PR #89, opened 2022-03-07
Start a demo of clicking a road and plugging it into osm2lanes

dabreegster:
The https://a-b-street.github.io/osm2lanes/ web interface is great, but an easier user flow is to just click a map and automatically grab the way from Overpass. This PR does the first half of that -- click a road in Mapbox, find the nearest way, display the road on the map (just as a line-string).

The second half will actually plug into osm2lanes; I'll resume #60 for that.

I'm doing this all in vanilla JS, no frameworks or anything. The goal is to make example code for other people to build off of, hopefully in time for FOSSGIS this weekend.

### DIFF
--- a/web/serve_locally.py
+++ b/web/serve_locally.py
@@ -1,0 +1,14 @@
+#!/usr/bin/env python3
+# This serves the current directory over HTTP. We need something more than
+# `python3 -m http.server 8000` to inject the CORS header.
+
+from http.server import HTTPServer, SimpleHTTPRequestHandler, test
+import sys
+
+class CORSRequestHandler (SimpleHTTPRequestHandler):
+    def end_headers (self):
+        self.send_header('Access-Control-Allow-Origin', '*')
+        SimpleHTTPRequestHandler.end_headers(self)
+
+if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


dabreegster:
Oh yeah, the demo:

https://user-images.githubusercontent.com/1664407/157035652-34f75c27-d752-4f05-a7f9-9f0775cf533d.mp4

web/serve_locally.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python3
```
dabreegster, line 1: I gave an attempt at using some modern workflow for web dev, but got overwhelmed. So for the moment, no bundling and stuff needed; the .html just pulls in all JS and CSS dependencies from a CDN. And to just run locally, this is my hack. If there's an npm / yarn / whatever tutorial that's easy to follow and would just let me bring in a few dependencies and not all of React, happy to follow that!

droogmic: its simple and it works.. I have not complaints.

I'll look into something structural, but only after the talk

droogmic (approved):

dabreegster:
Thanks for the review, I'll merge.

Getting JS / Rust interop through wasm-bindgen is way harder than I expected. You can't even pass around a `Vec<String>` apparently. But there's some serde magic to/from `JsValue`s that I'm looking into.

PR #95, opened 2022-03-12
Rust: Add Max Speed

droogmic:
I am pushing towards supporting something like this: https://www.rovz.nl/sites/zl-rovz/files/rovz_infra_ehkfoldeng.pdf
for the NL locale.

(It may no longer be up to date, but maxspeed-dependent separators is something we need to be able to support)


rust/osm2lanes/src/metric.rs
```diff
@@ -0,0 +1,330 @@
...
+    where
+        I: Iterator<Item = Metre>,
+    {
+        Self(iter.map(|m| m.0).sum())
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum Speed {
+    Kph(f64),
```
dabreegster, line 44: An enum is harder for an external user to deal with. Why don't we convert everything to kph, like distance? Also I think it vastly simplifies all of the serialization code!

dabreegster (changes requested):

rust/osm2lanes/src/metric.rs
droogmic: A renderer may/will want to show the speed limit per lane? And this visualisation will need to know the locale of how this is displayed, and this is know by osm2lanes (and hopefully not so much by the rendered)?

I would have preferred that as a simpler approach as well

rust/osm2lanes/src/metric.rs
```diff
@@ -0,0 +1,330 @@
...
+        Self(self.0 + other.0)
+    }
+}
+impl std::ops::AddAssign for Metre {
+    fn add_assign(&mut self, other: Self) {
+        *self = Self(self.0 + other.0);
+    }
+}
+impl std::ops::Mul<Metre> for f64 {
+    // The division of rational numbers is a closed operation.
```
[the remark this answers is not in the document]
droogmic: huh, this was there before, but I dont remember writing it, probably a copy paste?

```suggestion
```

rust/osm2lanes/src/metric.rs
droogmic: I just realised I should have documented my whole attempt at making this an unisgned integer.

but then I discovered the amazing world of US fractional speed limits :/

dabreegster: Good point, locale does matter for rendering. I wonder if it'd be simpler to give people a way to format a `Speed` struct with the locale for that use case. It makes the cross-language API bindings harder, though, since we're exposing methods on more structs.

I'd still vote for canonical units, but this approach is fine by me too

dabreegster (approved):

rust/osm2lanes/src/metric.rs
droogmic: I guess most renderers will need to know how to render the signs differently anyway, so they could convert back to mph?

idk

droogmic: maybe a struct like 

```rust
Speed {
    unit: Unit, # e.g. "mph"
    kph: f64,
}

Comment (droogmic) [reply rust/osm2lanes/src/metric.rs:43]: I think I will leave like this now indeed.

It might be hard (if they use the enum, not the flattened serialized API), but at least it has few surprised as it mirrors OSM's maxspeed tag.

Comment (dabreegster) [reply rust/osm2lanes/src/metric.rs:43]: Interesting idea. Why's that need to be repeated for every Speed value though? Why not just do `locale.speed_unit() -> Unit`?

Comment (droogmic) [reply rust/osm2lanes/src/metric.rs:43]: oh yeah, true.

I guess the the speed unit (along with driving side, and other locale details), would sit at the same level as highway and lifecycle does, outside the lanes.

Comment (droogmic) [reply rust/osm2lanes/src/metric.rs:43]: then you can just do `speed: f64` and have the data consumer use the global locale information...

I would just need to check that there isn't some edge case where cars have a speed limit in mph, but busses, trams, or bicycles have one in kph for example... that would suck.

File: rust/osm2lanes/src/metric.rs [hunk 3]
```diff
@@ -0,0 +1,329 @@
...
+    fn sum<I>(iter: I) -> Self
+    where
+        I: Iterator<Item = Metre>,
+    {
+        Self(iter.map(|m| m.0).sum())
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum Speed {
```
droogmic, line 42:
```suggestion

/// Vehicle Speed, used for speed limits and similar
// TODO: This part of the API may need to be revisited entirely
// It is unclear whether a speed unit is needed per lane,
// or if it could be provided for the entire road.
// Data consumers may also want to km/h value
// regardless of the unit used in the highway's locale
// in order to do distance over time calculations.
// The serialization of this is clunky and leaves a lotto be desired.
#[derive(Copy, Clone, Debug, PartialEq)]
pub enum Speed {
```

PR #96, opened 2022-03-13
README and Rust Toolchain Updates

droogmic:
Help guide newcomers more clearly.


dabreegster (approved):
There's always this tension between providing build instructions TLDR upfront and actually going into detail. I like this approach.

PR #98, opened 2022-03-14
Rust: Retrieve Locale from Overpass

droogmic:
image

### DIFF
--- a/rust/osm2lanes/src/overpass.rs
+++ b/rust/osm2lanes/src/overpass.rs
@@ -1,6 +1,7 @@
 use serde::Deserialize;
 
 use crate::tag::Tags;
+use crate::{DrivingSide, Locale};
 
 #[derive(Debug, Clone, Deserialize)]
 struct OverpassResult {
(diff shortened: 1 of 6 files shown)


rust/osm2lanes/src/overpass.rs
```diff
@@ -41,3 +46,69 @@ pub async fn get_way(id: ElementId) -> Result<Tags, reqwest::Error> {
...
+    let locale = Locale::builder()
+        .driving_side(driving_side.map_or(DrivingSide::Right, |d| d.parse().unwrap()))
+        .iso_3166_option(iso3166_2.or(iso3166_1))
+        .build();
+
+    let way_element = {
+        let mut elements = resp.elements;
+        elements.truncate(1);
+        elements.pop().unwrap()
+    };
```
droogmic, line 109: suggestions on how to better consume a vector and just return the first item are welcome...

dabreegster: Hmm. Performance is unlikely to matter here since we're not calling it in a tight loop (and networking lag dominates by far); why not just `resp.elements.remove(0)` and pay the cost of shifting the vector down? Alternatively, `swap_remove` is a good idea from https://www.reddit.com/r/learnrust/comments/hwseol/whats_the_idiomatic_way_to_move_the_first_element/

dabreegster (approved):
Super cool! Overpass can do some approximation of everything :)

rust/osm2lanes/src/overpass.rs
droogmic: yeah, so nothing significantly better than what I did for readability. I will leave it as is, thanks

PR #99, opened 2022-03-15
Rust: Fix Website Warnings

droogmic:
https://a-b-street.github.io/osm2lanes/

the roundtrip tags are shown again

warnings are deduplicated

### DIFF
--- a/rust/osm2lanes/src/transform/lanes_to_tags.rs
+++ b/rust/osm2lanes/src/transform/lanes_to_tags.rs
@@ -16,6 +16,12 @@ pub struct LanesToTagsConfig {
     pub check_roundtrip: bool,
 }
 
+impl LanesToTagsConfig {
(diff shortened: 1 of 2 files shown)


rust/osm2lanes/src/transform/lanes_to_tags.rs
```diff
@@ -16,6 +16,12 @@ pub struct LanesToTagsConfig {
     pub check_roundtrip: bool,
 }
 
+impl LanesToTagsConfig {
+    pub fn new(check_roundtrip: bool) -> Self {
+        LanesToTagsConfig { check_roundtrip }
+    }
+}
+
```
droogmic, line 24: needed due to `#[non_exhaustive]`, but if this gets much longer it will probably want a builder...

dabreegster (approved):

PR #102, opened 2022-03-19
Rust Refactor Separators

droogmic:
This became a bit big, sorry...

### DIFF
--- a/rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
+++ b/rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
@@ -134,8 +134,7 @@ struct Width {
 #[derive(Clone, Default, Debug)]
 pub struct LaneBuilder {
     r#type: Infer<LaneType>,
-    // TODO: is it better to have this be the direction relative to the way, or to the side.
-    // i.e. should a lane on the backward side be typically of the forward or backward direction
+    // note: direction is always relative to the way
     direction: Infer<LaneDirection>,
     designated: Infer<LaneDesignated>,
     width: Width,
--- a/rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
+++ b/rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
@@ -18,7 +18,7 @@ mod parking;
 use parking::parking;
 
 mod separator;
-use separator::insert_separators;
+use separator::{lane_to_edge_separator, lanes_to_separator};
 
 mod non_motorized;
 use non_motorized::non_motorized;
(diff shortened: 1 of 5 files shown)


rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -117,22 +117,25 @@ impl std::convert::From<LaneBuilderError> for RoadError {
...
     target: Infer<Metre>,
     max: Infer<Metre>,
 }
 
-#[derive(Default)]
-struct LaneBuilder {
+#[derive(Clone, Default, Debug)]
+pub struct LaneBuilder {
     r#type: Infer<LaneType>,
+    // TODO: is it better to have this be the direction relative to the way, or to the side.
```
dabreegster, line 137: The direction should be relative to the way, matching https://wiki.openstreetmap.org/wiki/Forward_%26_backward,_left_%26_right. If we did the alternative... the leftmost and rightmost lane could be reasoned about, but what about the middle lane on a 3 lane road? Meaningless to say it belongs to the left or right side of the road

rust/osm2lanes/src/transform/tags_to_lanes/separator.rs
```diff
@@ -3,111 +3,88 @@ use celes::Country;
...
-                color: None,
-                width: Some(Marking::DEFAULT_WIDTH),
-            }]),
-        }),
-        [_, Lane::Travel {
-            designated: LaneDesignated::Foot,
-            ..
-        }] => Some(Lane::Separator {
+    let [inside, outside] = lanes;
+    let direction_change = match [inside.direction.some(), outside.direction.some()] {
```
dabreegster, line 23: I was going to say, the rest of the PR is a bit complex 😅, but this code definitely got more readable - nice

dabreegster (approved):
@BudgieInWA, feel free to review PRs too, it could be a good way to pick up some Rust idioms.

Sorry for the slow review, I'm nearly finished with March and can properly work on this again...

rust/osm2lanes/src/transform/tags_to_lanes/separator.rs
droogmic: it would not have been worth it without something getting a bit better :)

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -117,22 +117,25 @@ impl std::convert::From<LaneBuilderError> for RoadError {
...
+    // i.e. should a lane on the backward side be typically of the forward or backward direction
     direction: Infer<LaneDirection>,
     designated: Infer<LaneDesignated>,
```
droogmic, line 140:
```suggestion
    r#type: Infer<LaneType>,
    // note: direction is always relative to the way
    direction: Infer<LaneDirection>,
    designated: Infer<LaneDesignated>,
```

Issue #103, opened 2022-03-19
Country Metadata


droogmic:
I think it would be good to share country-related metadata with other tools.
For example, we could fork https://github.com/streetcomplete/StreetComplete/blob/master/res/country_metadata into a different repo, and then include it both here and in streetcomplete as a compile time dependency to determine locale specific data.

Complication: https://github.com/streetcomplete/StreetComplete/blob/master/COPYING this license is incompatible with ours as far as I know.
Either we need to rewrite this from scratch, or discuss licensing with the StreetComplete maintainer.

dabreegster:
@westnordost I think you mentioned this idea elsewhere before. As a strawman proposal, what if we forked a streetcomplete/country_metadata repo with these YAMLs? And any opinions about an Apache license for that?

westnordost:
I looks like I missed this. Hm that's right, there is no mention of a license for that data, which by default makes it GPL-3.

I am willing to relicense that as CC-0 / public domain, but I am not the only (though almost, IIRC) person to have contributed to that, so the other people who have contributed to each file need to be asked first (and their contribution axed if they don't agree).
Apache License is more for software, is it not? This is just a data collection, so a CC-license seems to be more fitting. The more restrictive a license, the less contributors will likely have a problem with that. Would e.g. CC-BY-SA be acceptable?

westnordost:
I see @droogmic renamed the centerLineStyle.yml to center_line_style.yml. Is there a good reason for this or just a matter of taste? If there is a good reason for it, I could change the naming convention for all the files when putting the data CC0 licensed into a different repo

westnordost:
Alrighty, all the country metadata is now here: 

https://github.com/streetcomplete/countrymetadata/

Note that they are grouped by property (e.g. isLeftHandTraffic.yml and not US.yml) because it is easier to manage and contribute this way. For using the data, the other way around (US.yml etc.) is likely more convenient, which is why a build script in StreetComplete switches this around on downloading it. Maybe you would like to use something like that too:
https://github.com/streetcomplete/StreetComplete/blob/master/buildSrc/src/main/java/GenerateMetadataByCountry.kt

droogmic:
> I see @droogmic renamed the centerLineStyle.yml to center_line_style.yml. Is there a good reason for this or just a matter of taste? If there is a good reason for it, I could change the naming convention for all the files when putting the data CC0 licensed into a different repo

[...]

PR #104, opened 2022-03-19
Rust: Use StreetComplete Data Example

droogmic:
See #103

Do not merge until licensing issue is resolved.


droogmic:
Declining to replace with structural content

PR #105, opened 2022-03-20
Rust benchmark

droogmic:
I don't think performance will be an issue, but If you don't test it, you don't know.
violin


rust/osm2lanes/src/lib.rs
```diff
@@ -20,4 +20,5 @@ pub mod overpass;
 pub mod transform;
 pub use transform::{lanes_to_tags, tags_to_lanes, LanesToTagsConfig, TagsToLanesConfig};
 
-mod test;
+#[cfg(feature = "overpass")]
+pub mod test;
```
droogmic, line 24: ignoring the typo, I think this is the most idiomatic way to share test interfaces?

```suggestion
#[cfg(feature = "tests")]
pub mod test;
```

dabreegster (approved):
Nice. I've never seen criterion in use before, looks powerful

PR #106, opened 2022-03-21
Create WASM bindings and wire up into the vanilla JS Mapbox demo

dabreegster:
I believe this is the minimal work needed to call the Rust library from JS. It appears to work:
Screenshot from 2022-03-21 10-27-08

Still TODO: passing more options in, error handling, and an NPM package

### DIFF
--- a/rust/osm2lanes/src/lib.rs
+++ b/rust/osm2lanes/src/lib.rs
@@ -22,3 +22,8 @@ pub use transform::{lanes_to_tags, tags_to_lanes, LanesToTagsConfig, TagsToLanes
 
 #[cfg(feature = "tests")]
 pub mod test;
+
+#[cfg(target_arch = "wasm32")]
+mod wasm;
+#[cfg(target_arch = "wasm32")]
+pub use wasm::*;
--- a/web/index.html
+++ b/web/index.html
@@ -48,7 +48,11 @@
     <textarea id="input" rows="10" cols="80" disabled>Input</textarea>
     <textarea id="output" rows="10" cols="80" disabled>Output</textarea>
 
-    <script>
+    <script type="module">
+      import init, { js_tags_to_lanes } from "./pkg/osm2lanes.js";
+
+      await init();
+
       mapboxgl.accessToken =
         "pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpejY4NXVycTA2emYycXBndHRqcmZ3N3gifQ.rJcFIG214AriISLbB6B5aw";
       const map = new mapboxgl.Map({
@@ -113,6 +117,17 @@
 
         // Add the first feature to the map
         map.getSource("road").setData(geojson["features"][0]);
+
+        const input = {
+          key_values: geojson["features"][0]["properties"],
+          drive_on_right: true,
+        };
+        let lanes = js_tags_to_lanes(input);
+        document.getElementById("output").innerText = JSON.stringify(
+          lanes,
+          null,
+          2
+        );
       }
 
       map.on("click", (e) => {
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,3 +71,16 @@ trunk serve
 - The web demo is updated with every push on `main`, [see Workflow](./.github/workflows/web.yml)
 - The html website is part of the rest implementation at [`/rust/osm2lanes-web` ](./rust/osm2lanes-web)
 
+### JS bindings
+
+You can call the Rust library in JavaScript through WASM bindings.
+
+```shell
+cd rust/osm2lanes
+wasm-pack build --dev --target web -- --features wasm
+cd ../../web
+ln -s ../rust/osm2lanes/pkg/ .
+./serve_locally.py
+```
+
+Then open the link in your web browser and click a road.
--- a/rust/osm2lanes/Cargo.toml
+++ b/rust/osm2lanes/Cargo.toml
@@ -8,24 +8,28 @@ authors = [
 ]
 license = "Apache 2.0"
 
+[lib]
+crate-type = ["cdylib", "lib"]
+bench = false
+
 [dependencies]
 celes = "2.1"
+console_error_panic_hook = { version = "0.1.7", optional = true }
 log = "0.4"
 reqwest = { version = "0.11", optional = true, features = ["blocking", "json"] }
 serde = { version = "1", features = ["derive"] }
 serde_yaml = "0.8"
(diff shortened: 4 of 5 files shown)


dabreegster:
Ah right, my reference was https://rustwasm.github.io/wasm-bindgen/reference/arbitrary-data-with-serde.html

rust/osm2lanes/src/lib.rs
```diff
@@ -21,3 +21,8 @@ pub mod transform;
 pub use transform::{lanes_to_tags, tags_to_lanes, LanesToTagsConfig, TagsToLanesConfig};
 
 mod test;
+
+#[cfg(target_arch = "wasm32")]
```
dabreegster, line 25: I think another crate is overkill; it's not too much work to feature-gate a few dependencies and this modul

web/index.html
```diff
@@ -48,7 +48,11 @@
     <textarea id="input" rows="10" cols="80" disabled>Input</textarea>
     <textarea id="output" rows="10" cols="80" disabled>Output</textarea>
 
-    <script>
+    <script type="module">
+      import init, { js_tags_to_lanes } from "./pkg/osm2lanes.js";
+
+      await init();
```
dabreegster, line 54: I'm not convinced this is correct; I think everything here needs to be in an `async function`. But it kind of works out as long as the user doesn't click a road too quickly before this happens to finish

BudgieInWA: I think [top level await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#top_level_await) is a thing (though I personally haven't used module scripts directly in the browser before). So this should work, but at the cost of queuing the rest of the setup work behind `init()`.

You could kick off `init()` here, store the promise, and `await` it just before you use the lib (on line 120).

rust/osm2lanes/src/lib.rs
droogmic: maybe, I still think the separation helps in readability and separation of concerns?

I will do that work as part of https://github.com/a-b-street/osm2lanes/pull/60, if I still think it is needed.
It also has the added benefit of being a home for node js based integration tests

droogmic (approved):
Looks OK, just need to figure out how to integrate this into the deployed website.

web/index.html
dabreegster: Hmm, then we need extra state to avoid awaiting the promise multiple times. Cool that top-level await works directly though! I'll keep this in mind if we hit issues later

dabreegster:
Rebasing...

> maybe, I still think the separation helps in readability and separation of concerns?

Sounds good to me. I'm still not sure what the node config looks like, or how to wire up end-to-end JS integration tests. This is just a small step in that direction

web/index.html
BudgieInWA: FYI, awaiting the promise multiple times is also safe, at the cost of yielding to the event loop (like `setTimeout( _, 0)`).

BudgieInWA:
This breaks `trunk serve` in the rust dir for me:

```
~/c/osm2lanes  ➦ 098d9b7  rust  git co main
Previous HEAD position was 098d9b7 Rust benchmark (#105)
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
 ~/c/osm2lanes   main  rust  trunk serve
Mar 24 21:52:05.691  INFO 📦 starting build
Mar 24 21:52:05.692  INFO spawning asset pipelines
Mar 24 21:52:05.846  INFO copying & hashing css path="osm2lanes-web/assets/css/main.css"
Mar 24 21:52:05.846  INFO building osm2lanes-web
Mar 24 21:52:05.847  INFO finished copying & hashing css path="osm2lanes-web/assets/css/main.css"
   Compiling osm2lanes v0.1.0 (/home/ben/code/osm2lanes/rust/osm2lanes)
error[E0433]: failed to resolve: use of undeclared crate or module `wasm_bindgen`
 --> osm2lanes/src/wasm.rs:4:5
  |
4 | use wasm_bindgen::prelude::*;
  |     ^^^^^^^^^^^^ use of undeclared crate or module `wasm_bindgen`

error[E0432]: unresolved imports `crate::tags_to_lanes`, `crate::DrivingSide`, `crate::Locale`, `crate::TagsToLanesConfig`
 --> osm2lanes/src/wasm.rs:7:13
  |
7 | use crate::{tags_to_lanes, DrivingSide, Locale, TagsToLanesConfig};
  |             ^^^^^^^^^^^^^  ^^^^^^^^^^^  ^^^^^^  ^^^^^^^^^^^^^^^^^

error: cannot find attribute `wasm_bindgen` in this scope
  --> osm2lanes/src/wasm.rs:15:3
   |
15 | #[wasm_bindgen]
   |   ^^^^^^^^^^^^

error[E0433]: failed to resolve: use of undeclared crate or module `console_error_panic_hook`
  --> osm2lanes/src/wasm.rs:17:5
   |
17 |     console_error_panic_hook::set_once();
   |     ^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `console_error_panic_hook`

error[E0433]: failed to resolve: use of undeclared type `JsValue`
  --> osm2lanes/src/wasm.rs:38:5
   |
38 |     JsValue::from_serde(&lanes).unwrap()
   |     ^^^^^^^ use of undeclared type `JsValue`

error[E0412]: cannot find type `JsValue` in this scope
  --> osm2lanes/src/wasm.rs:16:31
   |
16 | pub fn js_tags_to_lanes(val: &JsValue) -> JsValue {
   |                               ^^^^^^^ not found in this scope

error[E0412]: cannot find type `JsValue` in this scope
  --> osm2lanes/src/wasm.rs:16:43
   |
16 | pub fn js_tags_to_lanes(val: &JsValue) -> JsValue {
   |                                           ^^^^^^^ not found in this scope

Some errors have detailed explanations: E0412, E0432, E0433.
For more information about an error, try `rustc --explain E0412`.
error: could not compile `osm2lanes` due to 7 previous errors
Mar 24 21:52:06.590 ERROR ❌ error
error from HTML pipeline

Caused by:
    0: error from asset pipeline
    1: error during cargo build execution
    2: cargo call returned a bad status
Mar 24 21:52:06.591  INFO 📡 serving static assets at -> /
Mar 24 21:52:06.591  INFO 📡 server listening at 0.0.0.0:8080
```

dabreegster:
Argh, I failed to test everything. Feel free to revert temporarily to unblock working on that, if you need to. Work and travel stuff mean I probably can't touch anything in osm2lanes until after April 6 or so. :\

PR #110, opened 2022-03-24
Rust Fix WASM

droogmic:


droogmic:
cc: @BudgieInWA

dabreegster (approved):
Separate crate makes sense after all. Thanks for cleaning up my mess! I need to actually configure the git hooks on my end. (I'm still used to envs where the presubmit tests are enforced and the commit won't go through)

droogmic:
> Separate crate makes sense after all. Thanks for cleaning up my mess! I need to actually configure the git hooks on my end. (I'm still used to envs where the presubmit tests are enforced and the commit won't go through)

I am disappointed we (software engineering) don't have better standards for sharing pre-commit/pre-push hooks :(

Issue #101, opened 2022-03-19
Allow configuring how and when osm2lanes guesses


BudgieInWA:
My idea for surfacing the provenance of values in the output to the user, and providing clarity and control for the round-trip problem is this: **Let the user configure if and when osm2lanes should guess details**, instead of annotating every detail with its provenance.

### Provides a straight forward way for the user to get what they want

If osm2lanes always guesses and annotates data, the chance that the user needs to process the output to filter out guesses that they don't want is high. If the level of guessing is customisable, it is likely that one of the options is exactly what they want:

- **Strict Mode** for people implementing an editor and round-tripping: describes exactly what the tags describe, warnings for any tag that could not be reconciled, and errors for straight-up conflicts.
- **Consensus Mode** for people rendering an "accurate" map (and probably for the visualisation in an editor too): makes assumptions that mappers expect, such as those documented on the wiki.
- **Fanciful Mode** for people making something pretty or A/B Street: assumes as much as possible, describing a plausible road that is not proven *wrong* but the tags, and never fails.
- **Custom Mode** could let the user opt into and even configure exactly the guesses they want.

### Keeps the schema simple

Tagging every value in every entry in the output will turn the resulting JSON into a wall of text in no time, and I am not looking forward to any part of that experience.

### Provides a way to check provenance

The same tags can be converted in different modes, and the user can compare between the two outputs to make more nuanced decisions.

### Enables flexibility in implementing guesses

I feel like this will be easier to implement than tracking provenance, especially if we come up with guesses that have nock-on effects in the rest of the parsing process (such as arbitrarily resolving conflicts). There will be no need to keep track of all the values that have been effected by the guess, while the user can still figure out what was assumed.

dabreegster:
In case the github reactions are insufficient -- I'm in agreement here. The current schema is slowly getting complex. The common case will have loads of inferred values for width!

Issue #72, opened 2022-02-08
Fix Lane Logic on Website


droogmic:
On the current website (https://a-b-street.github.io/osm2lanes/) you will notice that a tag of `lanes=6` becomes `lanes=5`!

This is clearly a bug.

Start by making a test case that fails:
https://github.com/a-b-street/osm2lanes/blob/main/data/tests.yml

And then try to find out how to fix it!

droogmic:
Thanks @BudgieInWA

PR #111, opened 2022-03-25
Fix lane count calculation - use Scheme structs to avoid re-reading/re-interpreting tags

BudgieInWA:
I am submitting this draft PR to get input on the new approach to `driving_lane_directions`, rust style and `tests.yml`. This PR is aiming towards ~~#69~~ #72.

### Are we happy with this overall structure for `driving_lane_directions`? 
bf5df9d8abdad9b72b4a5b64d558784d6695d1e8 shows the concise version, and dc5de27a3ee539cf25735219d2fdd2c8ffc7e677 shows how useful tag validation warnings can be added (thinking about #101 and how an editor would like to use `tags_to_lanes`).

### Any feedback on my rust style?
I want to fit into this project, more than anything else, but couldn't help trying out `is_some_and` and using lots of `unwrap_or`.

### How do the roundtrip tests work?
I couldn't figure it out just yet. Some guidance in the docs about writing and running `tests.yml` tests would be super awesome. (I figured out `cargo test --feature tests`, which I'll add to the docs if I get there first.)


data/tests.yml
```diff
@@ -1125,12 +1125,37 @@
...
+  - type: shoulder
+
 - way_id: 335668924
+  description: looks like a 1 lane wide alley with no markings
   tags:
     highway: 'road'
     lanes: '1'
     sidewalk: 'no'
   driving_side: right
+  # I don't think this expected output is correct... -- BudgieInWA
```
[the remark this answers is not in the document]
droogmic: feel free to fix the test and then either disable it for all languages or fix it in code.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -566,58 +568,75 @@ pub fn tags_to_lanes(
...
-            1
-        } else {
-            2
+        .and_then(|num| num.parse::<usize>().ok());
+    let tagged_both_ways = tags
+        .get("lanes:both_ways")
+        .and_then(|num| num.parse::<usize>().ok());
+    let num_both_ways = tagged_both_ways.unwrap_or(0);
+    let tagged_bus_lanes = tags
+        .get("lanes:bus")
```
droogmic, line 583: comment for myself: I think we look at this tag again in busway processing, need to see if this is a code smell or something inherent to the problem.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -566,58 +568,75 @@ pub fn tags_to_lanes(
...
+            warn!("lanes:backwards tag on a oneway road")
+        }
+        if tagged_both_ways.is_some() {
+            warn!("lanes:both_ways tag on a oneway road")
+        }
+        match (tagged_lanes, tagged_forward) {
+            (Some(l), None) => (l, 0),
+            (Some(l), Some(f)) => {
+                if l != f {
+                    warn!("lanes tag does not agree with lanes:forward on oneway road")
```
droogmic, line 598: we have a warnings Vec to structurally return this kind of information.
As we are a low level library, we need to give full control to the client on what they want to do with warnings

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -566,58 +568,75 @@ pub fn tags_to_lanes(
...
+            (Some(l), Some(f), None) => (f, l - f - num_both_ways),
+            (Some(l), None, Some(b)) => (l - b - num_both_ways, b),
+            (Some(1), None, None) => {
+                warn!("lanes=1 on a two way road");
+                (1, 1) // assume they meant lanes=2, I guess
+            }
+            (Some(l), None, None) => {
+                // Try to divide the lanes equally.
+                if (l - num_both_ways) % 2 != 0 {
+                    warn!("lanes - lanes:both_ways cannot be evenly distributed");
```
droogmic, line 625: hmmmmmmm
@dabreegster I think you made the decision to round in the favour of forward direction.
Do you think that is consistent enough to not raise a warning?

droogmic (changes requested):
Looks good, thanks for the work so far.

You are working on issue #72

droogmic:
> but couldn't help trying out is_some_and and using lots of unwrap_or

Looks pretty idiomatic to me, but then again I am not expert.

droogmic:
> I figured out cargo test --feature tests

This was a bug, it should be fixed now.

You should be able to run all the tests with just `cargo test`

The roundtrip works by taking the output (only the output), and checking that you can get back to the same spec by going to tags and back again (round trip).

Why not tags -> spec -> tags?
Well: spec to tags is a one to many mapping, so this shouldn't guarantee a roundtrip.

Documentation is very welcome!

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
dabreegster: A warning sounds appropriate. Unless the OSM wiki is very clear what this case means, it's a case of missing / ambiguous data. Good catch!

rust/osm2lanes/src/lib.rs
```diff
@@ -4,6 +4,8 @@
 //! WARNING: The output specification and all of this code is just being prototyped. Don't depend
 //! on anything yet.
 
+#![feature(is_some_with)]
```
dabreegster, line 7: This would require nightly rust to build, right? Since we're already requiring it for formatting, should we just make use of it everywhere?

(I'd push back a little -- I think it would require all downstream consumers to use nightly too. It's a pity, since there are features like this that're quite handy)

dabreegster (commented):
Thanks @droogmic for reviewing, feel free to take the lead on this. I may have some time over the weekend to catch up, but other stuff first...

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
droogmic: https://wiki.openstreetmap.org/wiki/Key:lanes#Extended_tagging

I agree with the warning, no change needed.

rust/osm2lanes/src/lib.rs
droogmic: why nightly?

dabreegster: ``#![feature]` may not be used on the stable release channel`
https://github.com/rust-lang/rust/issues/93050
Until this is stabilized, I think this forces us to use nightly rust

droogmic: it seems features like that are unlikely to ever land in stable, I am actually surprised that one went in:
https://github.com/rust-lang/rust/pull/75298
we can go through the cases where it was used here and replace it with something slightly more verbose :)

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -566,58 +568,75 @@ pub fn tags_to_lanes(
...
-        0
     } else {
-        1
-    };
-    (num_driving_fwd, num_driving_back)
+        match (tagged_lanes, tagged_forward, tagged_backward) {
+            (_, Some(f), Some(b)) => {
+                if tagged_lanes.is_some_and(|&l| l != f + b + num_both_ways) {
+                    warn!("lanes tag does not agree with lanes:forward, lanes:backward and lanes:both_ways");
+                }
```
droogmic, line 613:
```suggestion
                if let Some(num_lanes) = tagged_lanes {
                    if num_lanes != f + b + num_both_ways {
                        warn!("lanes tag does not agree with lanes:forward, lanes:backward and lanes:both_ways");
                    }
                }
```

data/tests.yml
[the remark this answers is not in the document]
BudgieInWA: One travel lane with `direction: both` seems sensible to me, I will go with that.

I guess we need to decide how to represent the full continuum of unmarked roads, from narrow alleys where no passing is possible, through "1.5 lane" roads where the shoulder is used for passing, to residential streets with no lane markings wide enough for cars parked on both sides *and* two cars to barely pass between them. Tagging seems all over the place (see the [Key:lanes talk page](https://wiki.openstreetmap.org/wiki/Talk:Key:lanes).

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
BudgieInWA: It makes sense to me, to look at it is both places if this function is trying to determine the number of lanes definitively. The `lanes:bus` tag is only being looked at because the way is under tagged, and we want to guess that there are some unrestricted lanes as well. Which would suggest that we should potentially look at `busway` tags here too.

rust/osm2lanes/src/lib.rs
BudgieInWA: I agree with staying on stable. I didn't know what `#![feature]` was, but IntelliJ suggested it, so I thought I'd get your input on it. Now I know!

I only used it one time, and can easily work around it.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
BudgieInWA: This does seem like the right place to generate most of these warnings, so I will upgrade all of the `warn!`s to instead feed into the `RoadWarnings` vec.

droogmic:
Agreed, the content and location of the warnings appears correct.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -565,59 +568,191 @@ pub fn tags_to_lanes(
...
+                                "lanes:backward",
+                            ])),
+                        });
+                    }
+                }
+                (f, num_both_ways, b)
+            }
+            (Some(l), Some(f), None) => (f, num_both_ways, l - f - num_both_ways),
+            (Some(l), None, Some(b)) => (l - b - num_both_ways, num_both_ways, b),
+            (Some(1), None, None) => (0, 1, 0),
```
droogmic, line 714: 1 lane becomes a centre lane? seems fishy?

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -565,59 +568,191 @@ pub fn tags_to_lanes(
...
+                        warnings.push(RoadMsg::Ambiguous {
+                            description: None,
+                            tags: Some(tags.subset(&[
+                                "lanes",
+                                "lanes:forward",
+                                "lanes:both_ways",
+                                "lanes:backward",
+                            ])),
+                        });
+                    }
```
droogmic, line 708: Can num_both_lanes could come from center_turn_lanes? Add to warning.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -565,59 +568,191 @@ pub fn tags_to_lanes(
...
+///
+/// Validates `lanes[:{forward,both_ways,backward}]=*` to determine the precise answer.
+/// Uses `lanes:{bus,psv}[:{forward,backward}]=*`, `busway[:{left,both,right}]=*` and
+/// `centre_turn_lane=yes` to make assumptions in the absence of precise tagging.
+fn driving_lane_directions(
+    tags: &Tags,
+    locale: &Locale,
+    oneway: Oneway,
+    warnings: &mut RoadWarnings,
+) -> (usize, usize, usize) {
```
droogmic, line 602: can we test this in isolation, it is quite complex now :)

droogmic (changes requested):
Looking good so far, surprisingly complex :)

BudgieInWA:
I am finding more and more out about `Tag`s library and the parsing techniques and good `RoadMsg` construction, still a little way to go. But, there is a fundamental problem:

### The problem with locking in a final guess on the lanes distribution at the start

We cannot assume or guess missing parts of the `lanes:{forward,backward}=` scheme without parsing everything else. See how 28300bc adds tons of code parsing other schemes. Definitely smelly.

We need a way to narrow down our guess as we go. Random thoughts:

- Maybe we use `Infer` like values too keep track of guesses.
- Maybe in guessing mode, start with the assumption that there is one normal lane in each direction, then *add lanes* instead of *mutating existing lanes*. Then at the end we can do a round of assuming (e.g. distribute remaining lanes).
- Maybe each tagging scheme has its own `struct` type using `Infer`ed values, which remains internally consistent, to help in the code which validates each tag schema, checks for conflicts, and wants to jump between them in order to do these complicated guesses. Then the validated and consistent schemes are used to build the road. Maybe having code that defines the test and guesses for each requirement between schemes is *exactly the right amount* of code (e.g.  `lanes:bus:forward=2` requires at least 2 forward lanes, or increases the guesses forward lanes by 2), but it still feels a bit like double handling (the whole point of the Builder pattern is to keep track of the jumping between different inputs, right?)
- It ties into validating: `psv:lanes=designated|` requires `lanes=2` to parse properly, so it implies lanes=2 (and is an error if lanes!=2)
- It feels a bit like Logic Programming where different tags are facts that work to collapse the possibility space.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
BudgieInWA: It is a bit fishy, but I think it is because describing "how many lanes in each direction does a skinny allyway have" is fishy. I might tag them "lane_markings=no width=5", but we have to represent that in schema somehow.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
BudgieInWA: I agree! (if this function ends up keeping all of its responsibilities.)

Does that mean a set of tests in tests.yaml, or rust tests?

droogmic:
> * with the assumption that there is one normal lane in each direction, then _add lanes_ instead of _mutating existing lanes_. Then at

I agree with everything you say, but I also don't really know the best way forward.

> but it still feels a bit like double handling (the whole point of the Builder pattern is to keep track of the jumping between different inputs, right?)

correct

> Maybe in guessing mode, start with the assumption that there is one normal lane in each direction, then add lanes instead of mutating existing lanes. 

I agree, that we should maybe try this instead. The downside is that all the other logic needs to account for this new system, so it will be a painful refactor :(.

A simplified version of the logic you have now should probably still exist, to populate a top level `Infer` field in roadbuilder that may get invalidated by clearer tags later (like your bus example)?

In any event, it sounds like you understand the problem. Now you just need to decide how hard you make the solution for yourself ;)

droogmic:
I added you to the project, so hopefully I dont need to approve the CI anymore

BudgieInWA:
> In any event, it sounds like you understand the problem. Now you just need to decide how hard you make the solution for yourself ;)

This was just the right thing to hear :) There are a couple of hacky things that I could try, but I don't see them coming together, so I decided to make the solution very hard for myself :D 

> A simplified version of the logic you have now should probably still exist, to populate a top level `Infer` field in roadbuilder that may get invalidated by clearer tags later (like your bus example)?

In [c9ca0eb](https://github.com/a-b-street/osm2lanes/pull/111/commits/c9ca0ebfee220521b2dd72e463dfee37ed3f4907) I started breaking out schema parsing from road building, so that the lanes schema can have access to the already parsed bus lanes. It looks like a direction that I want to keep exploring... which leads to the "hard" part:

Then [0496070](https://github.com/a-b-street/osm2lanes/pull/111/commits/04960707998117b9973ce319f7a5f78485c7a22f) I started writing the overall structure from scratch, as if I had access to `Hunch` and `Scheme` types, and I tried to plot a path through all the steps of assuming the value of some schema given the road, adding tags into the existing assumptions/requirements, etc.

I think `Hunch` needs breaking into two parts, as the comments hint at, but I am actually very happy with `Scheme` and how it helps be think about the problem:
1. Decide a canonical way to represent the scheme - e.g. all of the `busway` tags - in the context of `oneway` and `oneway:bus` can only boil down to: 'is there at least one lane" in each direction(and possibly if it is the opposite direction to expected, depending on interpretations of `opposite_lane`).
3. write `default(&Road) -> Busway` which looks at those dependant tags on the road
4. write `set_tags(mut self, &Tags)` which sets as much detail as possible (and possible makes other assumptions, even guesses), detects any conflicts with already recorded Implications (or should it just read the dependant schemes from the road again, to detect conflicts with implications)
5. write `guess` separately, with whatever crazy logic that I don't want anywhere near the strict-mode parsing path, etc.
6. write `impl Road { update_lanes_with(&mut self, scheme: Busway) {} }` based solely on the values that come out the the `Hunch`es in a schema.

As we think about it more, we can figure out what ordering requirements we need to enforce between all of the `Scheme` updating, and road building steps. It feels like there are comfortable places to put all of the pieces of complexity that I have come across so far: none of the schemes or the assumptions or guesses are that complicated themselves, and this way of dividing the responsibility's up seems to allow each of those things to be expressed in the right context. E.g. "if `lanes=` is not tagged, assume there is one lane in each travel direction (more on freeways) in addition to those described by all the bus tags" is the straightforward assumption that we want to make to support naked `busway` tags and the like. By deciding that busway is handled first, and implementing the assumption logic separate from the tag parsing logic, that becomes the [appropriately straightforward](https://github.com/a-b-street/osm2lanes/pull/111/files#diff-a8762abba191ab492eb7bcb0d196a929c60681561d371c85308a4c3b040e0c0dR180-R221) `let total_bus_lanes_forward = sum_bus_lanes(); ( Assume(default_forward + total_bus_lanes), ...)`.

I would love for @droogmic and/or @dabreegster to have a read through `road.rs` to
- see how close my attempts have approached your existing solution :P,
- see if the way `tags_to_lanes`, `Road`, `Scheme` and `Hunch` fit together in a way is comprehensible, and if so
- help decide what shape `Hunch` and some other types should have, and
- weigh in on how to approach implementing this idea further.

Next steps are
 - decide viability / approach: 
 	- rebuild: add 'strict mode output' to tests, so we can start from scratch and build up parsing of all schemes as written, then come back later to implement guesses (I am happy to get stuck into this kind of approach over some time)
 	- iterate: combine `Hunch` and `Infer`, change `RoadBuilder` to work with `Scheme`, then implement schemes one by one, including existing guessing (sounds sensible if the architectures are similar enough, and we are happy to sacrifice a few tricky tests to begin with)
 - implement type APIs and rust fiddlyness for Hunch and Scheme
 - modify existing RoadBuilder, or write a new one, to work with `Schemes`
 - rewrite all scheme parsing (!) in the new format (which might not be tooo horrible)

BudgieInWA:
I'm having too much fun exploring the rust type system to get anything compiling yet, but that is my next step. I should be able to get `Hunch`, `Theory` and `Scheme` compiling easily enough, to start actually seeing if they are useful.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -243,7 +250,7 @@ impl RoadBuilder {
         .take(num_driving_back)
         .collect();
         // TODO Fix upstream. https://wiki.openstreetmap.org/wiki/Key:centre_turn_lane
-        if tags.is("lanes:both_ways", "1") || tags.is("centre_turn_lane", "yes") {
+        for _ in 0..num_driving_both {
```
dabreegster, line 253: Do we have an example of multiple bidirectional lanes? I'm super curious how that would even work in practice O_o

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
dabreegster: Rust tests I'd imagine, since the input and output are quite specialized here. The unit tests can be in this file, since it's private

rust/osm2lanes/src/transform/tags_to_lanes/road.rs
```diff
@@ -0,0 +1,412 @@
...
+            Some(h) => h.strength(),
+            None => Guessed(()),
+        }
+    }
+}
+
+// trait HasProblems { fn iter<Problem>... }
+// type TheoryHistory = Vec<Problem>;
+struct Theory<'a, T> {
+    history: Vec<(Hunch<T>, Vec<&'a Theory<'a, T>>)>,
```
dabreegster, line 162: It's almost as if we're producing a full proof of how we arrive at the final answer and all of the assumptions made along the way. Really interesting idea, I've never seen something like this before. Interested to see where this approach goes

dabreegster (commented):
Sorry for the delay, I'm catching up on everything now (and feeling a little bit lost, but just because I've not looked at this in a few weeks)

I like the approach of teasing apart the schema for everything and having an explicit step where we combine/reconcile. I think it would also nicely handle the cases of lists like `psv:lanes = yes|yes|designated`, where we have an implied total number of lanes that might disagree with a total from another place.

`Hunch` seems like a natural extension of `Infer`. I'm curious where `Theory` will go; it feels complex to understand initially, but possibly very worth it.

rust/osm2lanes/src/transform/tags_to_lanes/road.rs
```diff
@@ -0,0 +1,412 @@
...
+    forward: Theory<'a, Count>,
+    backward: Theory<'a, Count>,
+    bothways: Theory<'a, Count>,
+    // directions: Theory<'a, Vec<LaneDirection>>,
+}
+
+struct Busway<'a> {
+    forward: Theory<'a, Option<LaneDirection>>,
+    backward: Theory<'a, Option<LaneDirection>>,
+}
```
droogmic, line 257: having so many lifetime specifiers usually indicates a code smell in rust FYI, as I don't see why you should need them.

rust/osm2lanes/src/transform/tags_to_lanes/road.rs
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    road.calculate_lanes_ltr().into()
+}
+
+/// A collection of schemas. We have a Hunch about each schema because we need at least an option,
+/// so why not track it as a Hunch?
+struct Road {
+    busway: Hunch<Busway>,
+    lanes: Hunch<Lanes>,
```
droogmic, line 33: how does this work? do we make assumptions about how lanes vs busways are arranged?

rust/osm2lanes/src/transform/tags_to_lanes/road.rs
```diff
@@ -0,0 +1,412 @@
...
+}
+
+/// A value of different certainties. Hunches combine together into another hunch in predefined ways.
+enum Hunch<T> {
+    // Unknown,
+    Tagged(T),
+    Implied(T),
+    Assumed(T),
+    Guessed(T),
+}
```
droogmic, line 80: I think a question you had was asking how to proceed in integrating this. I would start by adding and using this new type in a limited scope and then slowly refactoring out.

Trying to do it all at once will be difficult to make it pan out.

droogmic (commented):
Good ideas, but the scope has crept a bit ;)

droogmic:
At the moment my main concern is how you get such a big change integrated without needing to freeze all parallel development.

I think `Hunch`, and building lanes up from a base assumption of 1 (instead of converting them), are, if I understand correctly the 2 options that we can start with. I wonder to what extend we can deliver those now and expand to this more complete system over time. Things like `Theory` are probably something that can come later.

The reason I say this is that my current short term goal is to push quickly towards getting _most_ roads working in the countries that abstreet and similar tools usually operate, and I think the current simple schema can get us there without too much of an issue.

BudgieInWA:
> Good ideas, but the scope has crept a bit ;)
> I'm catching up on everything now (and feeling a little bit lost)...

I told you I had chosen the hard option to look at :) I will keep this idea/practice going in a branch and get back to the task at hand on this one. Thanks for watching me figure out rust.

I will start with pre-parsing the schemes like c9ca0eb, then build up to guessing `lanes=` last. I think it turns out that the order that we consider each scheme matters, even in my experiment. That might be the heart of it, and we can test it with just the schemes that `lanes=` depends on (anything that hints at specific kinds of lanes).

> At the moment my main concern is how you get such a big change integrated without needing to freeze all parallel development.
> ... Things like `Theory` are probably something that can come later.

Absolutely; I will merge main into here as needed once I go back to a compiling commit and we should only implement any ideas bit by bit (maybe we add what I was using in `Hunch` to `Infer`, maybe we add scheme structs with `::from(tags)`, etc.).

rust/osm2lanes/src/transform/tags_to_lanes/road.rs
BudgieInWA: We already do, and we need to because people want to tag just `busway:both=lane` to mean 4 lanes: a default 2 normal lanes with a busway on the outside, which we currently support with test [`way_id: 988980354`](https://github.com/a-b-street/osm2lanes/blob/620a83332cd696e08b41ce89e1d559ce7e1a386f/data/tests.yml#L815). It's why `busway` chooses `forward_inside()` vs `forward_outside()`.

rust/osm2lanes/src/transform/tags_to_lanes/road.rs
BudgieInWA: I agree, that's why I was happy to split any error tracking out of `Hunch`, to see what I wanted, which is `downgrade` and `reconcile`.

And thanks for being a persistent voice of reason :)

rust/osm2lanes/src/transform/tags_to_lanes/road.rs
BudgieInWA: Yep. I think something forces me to include it, and rust needs assurance here, but I haven't looked into it yet. Hopefully its not needed in the end!

rust/osm2lanes/src/transform/tags_to_lanes/road.rs
BudgieInWA: Yes, it just feels silly to throw any of the history away at any of the points where I have the option to. Iterating over the whole set of problems seems like the first useful reason to have any history around.

This line might be the reason I put in the lifetime. I am trying to keep a reference to the `Theory`/`impl TheoryLike` thing in the history, so we don't copy...

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
BudgieInWA: I don't really know. I have seen elsewhere the assumption that there can be at most 1 bothways lane (JOSM Lanes I think). 

The reason for the change is actually that the number of bothway lanes is more specifically what we need to look at here, as unlikely as the viability might be - e.g. it supports `lanes:both_ways=2` just fine, even if it breaks someone else's assumptions.

BudgieInWA:
@droogmic I started focusing on solving the original problem again :)

This is almost ready to merge. Probably a few rust-isms to tweak.

I plan on refactoring all of the bus lane tag parsing to use the `*Scheme` pattern that I have used here, but would prefer not to block this PR on those. As long as the direction that this PR takes `lanes_to_tags` is the right direction, I'm happy to merge, now that most of the tests are passing.

rust/osm2lanes/src/transform/tags_to_lanes/bus.rs
```diff
@@ -55,84 +70,101 @@ pub(super) fn bus(
...
-        "opposite_lane",
-    ) {
-        if tags.is("oneway", "yes") || tags.is("oneway:bus", "yes") {
-            // TODO: does it make sense to have a backward lane on the forward_side????
-            let lane = road
-                .forward_inside_mut()
+}
+
+impl RoadBuilder {
+    pub fn set_busway_scheme(
```
dabreegster, line 146: Yeah, this two-step approach of "figure out what's being described" and then "apply it to the road being built up" is way cleaner to read (and even unit test more granularly, if we ever wanted to)

rust/osm2lanes/src/transform/tags_to_lanes/bus.rs
```diff
@@ -55,84 +70,101 @@ pub(super) fn bus(
...
+        tags: &Tags,
+        locale: &Locale,
+        oneway: &Oneway,
+        warnings: &mut RoadWarnings,
+    ) -> Self {
+        let mut busway = Self {
+            forward_direction: Infer::Default(None),
+            backward_direction: Infer::Default(None),
+        };
+
```
dabreegster, line 84: The tiniest of style nitpicks: just one blank newline? cargo fmt may fix for you

rust/osm2lanes/src/transform/tags_to_lanes/lane.rs
```diff
@@ -0,0 +1,213 @@
...
+        lanes_bus: &LanesBusScheme,
+        _locale: &Locale,
+        warnings: &mut RoadWarnings,
+    ) -> Self {
+        // The tags for this schema (which we will validate).
+        let tagged_lanes: Option<usize> = tags.get_parsed(&LANES, warnings);
+        let tagged_forward: Option<usize> = tags.get_parsed(&(LANES + "forward"), warnings);
+        let tagged_backward: Option<usize> = tags.get_parsed(&(LANES + "backward"), warnings);
+        let tagged_bothways: Option<usize> = tags.get_parsed(&(LANES + "both_ways"), warnings);
+        let tagged_center_turn_lanes = match tags.is("centre_turn_lane", "yes") {
```
dabreegster, line 29: Should we emit a deprecation warning? https://wiki.openstreetmap.org/wiki/Key:centre_turn_lane

dabreegster (commented):
This is pretty fantastic!

data/tests.yml
```diff
@@ -794,6 +795,7 @@
 - description: bus:lanes:backward=designated
   skip_python: true
   skip_kotlin: true
+  rust: false # fails roundtrip (outputs busway:right=opposite_lane tag)
```
droogmic, line 798: Can you paste the full error here?
I don't particularly like taking a step backward in passing tests :P

data/tests.yml
```diff
@@ -1080,39 +1082,67 @@
...
+      direction: forward
+      designated: motor_vehicle
+    - type: travel
+      direction: forward
+      designated: motor_vehicle
+    - type: travel
+      direction: backward
+      designated: motor_vehicle
+    - type: travel
+      designated: foot
```
droogmic, line 1106: given the new indenting, refactor to the new test scheme:

```
  road:
    highway: road
    lanes:
    - type: travel
      designated: foot
    - type: travel
      direction: forward
      designated: motor_vehicle
    - type: travel
      direction: forward
      designated: motor_vehicle
    - type: travel
      direction: backward
      designated: motor_vehicle
    - type: travel
      designated: foot
```

droogmic: note the one additional layer of nesting.

it saves me doing it later :)

this applies to the other cases as well

rust/osm2lanes/src/tag/mod.rs
```diff
@@ -90,6 +91,23 @@ impl Tags {
         self.map.get(k.as_ref()).map(String::as_str)
     }
 
+    /// Get the value for the given key and parses it into T.
+    pub fn _get_and_parse<V: FromStr>(&self, key: &str) -> Option<V> {
```
droogmic, line 95: a `pub` function with an underscore prefix?
I think both can be removed and then it will be consistent.

rust/osm2lanes/src/test.rs
```diff
@@ -115,7 +115,7 @@ impl std::fmt::Display for TestCase {
 pub fn get_tests() -> Vec<TestCase> {
     let tests: Vec<TestCase> =
         serde_yaml::from_reader(BufReader::new(File::open("../../data/tests.yml").unwrap()))
-            .expect("invalid json");
+            .expect("invalid yaml in data/tests.yml");
```
droogmic, line 118: oops, thanks :D

rust/osm2lanes/src/transform/lanes_to_tags.rs
```diff
@@ -84,6 +84,54 @@ pub fn lanes_to_tags(lanes: &[Lane], locale: &Locale, config: &Config) -> TagsRe
...
+                lane,
+                Lane::Travel {
+                    designated: Designated::Motor,
+                    direction: Some(Direction::Both),
+                    ..
+                }
+            )
+        }) {
+            tags.checked_insert("lanes:both_ways", "1")?;
+            // TODO: add LHT support
```
droogmic, line 130: emit an unimplemented warning if you want to leave this as is?

rust/osm2lanes/src/transform/tags_to_lanes/bus.rs
```diff
@@ -19,6 +19,21 @@ impl LaneBuilder {
     }
 }
 
+const BUSWAY: TagKey = TagKey::from("busway");
+
+#[derive(Debug)]
+pub struct BuswayScheme {
+    pub forward_direction: Infer<Option<Direction>>,
```
droogmic, line 26: Without looking at the implementation, this is not super intuitive...

So the forward direction is either forward or backwards,
we can infer this to some level of confidence, including inferring that it is neither forward nor backwards?

I will now read the implementation...

rust/osm2lanes/src/transform/tags_to_lanes/bus.rs
```diff
@@ -55,84 +70,101 @@ pub(super) fn bus(
...
+                );
+            }
+        }
+        if tags.is(
+            BUSWAY + locale.driving_side.opposite().tag(),
+            "opposite_lane",
+        ) {
+            if tags.is("oneway", "yes") || tags.is("oneway:bus", "yes") {
+                // TODO: does it make sense to have a backward lane on the forward_side????
+                busway.forward_direction = Infer::Direct(Some(Direction::Backward));
```
droogmic, line 128: aha, this is the edge case.
So in australia (LHT) given a oneway `busway:right=opposite_lane`

https://wiki.openstreetmap.org/wiki/Key:busway
> Bus lanes in both directions, on left and right sides of a one-way road: [busway]():[left](https://wiki.openstreetmap.org/wiki/Key:left)=opposite_lane + [busway]():[right](https://wiki.openstreetmap.org/wiki/Key:right)=lane

I would expect a bus lane, as the furthest right lane, going in the backward direction (i.e. contraflow).

I struggle with what `busway.forward_direction = Direction::Backward`  means.

droogmic: I guess you are working around the fact that there is no backward_direction lanes to turn into bus lanes?

rust/osm2lanes/src/transform/tags_to_lanes/bus.rs
droogmic: I think this might just be an `Infer<bool>`

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -84,17 +90,17 @@ impl std::convert::From<Oneway> for bool {
 #[derive(Copy, Clone, Debug)]
 pub enum Infer<T> {
     None,
+    Guessed(T),
     Default(T),
```
droogmic, line 94: Maybe add some comments/examples of how Guessed differs from default?

I guess something like:

Guessed: no safe default value exists, so a potentially incorrect value was assumed in the absence of any relevant tags.
Default: A safe default value exists, and was used in the absence of any relevant tags.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -84,17 +90,17 @@ impl std::convert::From<Oneway> for bool {
...
+    Calculated(T),
     Direct(T),
 }
 
 impl<T> Infer<T> {
     pub fn some(self) -> Option<T> {
         match self {
             Self::None => None,
-            Self::Default(v) | Self::Direct(v) => Some(v),
+            Self::Guessed(v) | Self::Default(v) | Self::Calculated(v) | Self::Direct(v) => Some(v),
```
droogmic, line 103: I wonder if there are times when we don't want to use a guessed value (due to it's potentially incorrect non-default nature), but this should be fine for now :)

rust/osm2lanes/src/tag/mod.rs
droogmic: (you could use this function int he implementation of the below, otherwise I may suggest just removing it for now, as I don't know when we would use it over the below.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -578,59 +618,7 @@ pub fn tags_to_lanes(
...
-        } else {
-            2
-        }
-    } else if oneway.into() {
-        0
-    } else {
-        1
-    };
-    (num_driving_fwd, num_driving_back)
+impl Tags {
```
droogmic, line 621: empty impl?

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -546,15 +569,32 @@ pub fn tags_to_lanes(
...
-    let mut road: RoadBuilder = RoadBuilder::from(tags, locale, &mut warnings)?;
+    // Parse lane count schemas first.
+    let oneway =
+        Oneway::from(tags.is_any("oneway", &["yes", "-1"]) || tags.is("junction", "roundabout"));
+    let busway = BuswayScheme::new(tags, locale, &oneway, &mut warnings);
+    // let bus_lanes = BusLanes::from(tags, locale, &oneway, &mut warnings);
+    // TEMP: lets use the lanes:bus schema to summarise bus lanes for the lanes scheme.
+    let lanes_bus = LanesBusScheme {
+        forward: busway.forward_direction.map_or(0, |o| if o.is_some() { 1 } else { 0 }),
+        backward: busway.backward_direction.map_or(0, |o| if o.is_some() { 1 } else { 0 }),
```
droogmic, line 583: this construct seems fishy, probably due to comments above?

droogmic (changes requested):
Good progress, thanks.

rust/osm2lanes/src/transform/tags_to_lanes/bus.rs
BudgieInWA: I have my IDE setup to run cargo format, but it produces different results to `cargo fmt` on the command line (maybe due to the config specifying nightly?). Now I know to run `cargo fmt` before commiting.

droogmic: there is a rustfmt config file, and a toolchain file, that should correctly configure your rust setup. not sure why your IDE might ignore it.

rust/osm2lanes/src/transform/tags_to_lanes/bus.rs
BudgieInWA: This is from the existing logic. I never did understand it, perhaps it was working around oneway only having forward lanes?

It is probably right to say `backward_direction = Backward` in this case, but that depends on the interpretation of the scheme. I'd say so, and this isn't covered by tests, so I'll make the change. In the context of a `oneway=yes` road "opposite" must mean "backward": 

> Bus lanes in both directions, on left and right sides of a one-way road:
busway:left=opposite_lane + busway:right=lane [RHT]
https://wiki.openstreetmap.org/wiki/Key:busway#Examples

`busway.forward_direction` is better named "forward side direction", so `busway.forward_side_direction = Direction::Backward` is how we describe what the code was doing: `road.forward_inside_mut().direction = Infer::Direct(Direction::Backward)` (though my impl puts the `Direction::Backward` lane on the outside).

rust/osm2lanes/src/transform/tags_to_lanes/bus.rs
BudgieInWA: See my [other comment](https://github.com/a-b-street/osm2lanes/pull/111#discussion_r846609663) and new name `forward_side_direction`.

If the busway scheme can't describe `forward_side_direction = Backwards` then `forward: Infer<bool>` is better. What about LHT `busway:left=opposite_lane`?

I haven't tried to understand all the assumptions and usecases for busway, so feel free to dictate or even push improvements.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
BudgieInWA: Absolutely. Perhaps a `fn some_confident(self) -> Option<T>`. Ideally guesses taint everything they are used to calculate (with appropriate use of `downgrade` type logic), so that by the time we get to the end, we can decide how to render the result (or when we get to road building), depending on the requirements.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
BudgieInWA: Mm, it is a little. I am using `lanes_bus.backward` to count lanes on the backward side of the road (which `BuswayScheme` will switch direction if it needs to). I suppose it depends (if we even need to support such a thing) how `lanes:bus` would represent the contraflow bus lane on the unexpected side of the road situation.

rust/osm2lanes/src/transform/lanes_to_tags.rs
BudgieInWA: Got it.

data/tests.yml
BudgieInWA: I just get this now, which is different to before. I should be able to look into it later, as `set_busway_scheme` is supposed to work with the checks parsing `LanesScheme`. Perhaps more of those checks between schemes that I had as TODO will illuminate the problem... 

```
running 6 tests
test metric::speed::tests::test_speed ... ok
test locale::tests::test_locale ... ok
test metric::tests::test_speed ... ok
test tag::tests::test_tags ... ok
test test::tests::test_roundtrip ... FAILED
test test::tests::test_from_data ... ok

failures:

---- test::tests::test_roundtrip stdout ----
thread 'test::tests::test_roundtrip' panicked at 'called `Result::unwrap()` on an `Err` value: Msg(Unsupported { description: Some("no backward lanes for busway"), tags: None })', osm2lanes/src/test.rs:480:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

data/tests.yml
BudgieInWA: I think I have updated all the tests that will produce merge conflicts within the `output` lines.

BudgieInWA:
Now we just need to 

- solve our understanding of the busway scheme, at least so [the test passes](https://github.com/a-b-street/osm2lanes/pull/111#discussion_r846289902), and
- merge `main` in here.

@droogmic, I would be happy for you to do the merge, and arrange my new functions wherever you see fit. But if you prefer, I can do it when I have time during the week (my detours are the reason this turned into such a mammoth PR after all).

droogmic:
> Now we just need to
> 
>     * solve our understanding of the busway scheme, at least so [the test passes](https://github.com/a-b-street/osm2lanes/pull/111#discussion_r846289902), and
> 
>     * merge `main` in here.
> 
> 
> @droogmic, I would be happy for you to do the merge, and arrange my new functions wherever you see fit. But if you prefer, I can do it when I have time during the week (my detours are the reason this turned into such a mammoth PR after all).

to confirm, you want me to solve the merge conflict? it is indeed probably easier if I do it, given I caused it :P

data/tests.yml
droogmic: oh yeah, this is the the whole "we need a backward lane to turn into a busway problem".

can we modify the business logic in the edge case to switch from the outside backward lane to the inside forward lane in this case?

BudgieInWA:
> to confirm, you want me to solve the merge conflict? it is indeed probably easier if I do it, given I caused it :P

Yes, I would appreciate it.

droogmic:
The rust part is merged:
https://github.com/BudgieInWA/osm2lanes/pull/1

Some kotlin and python stuff is still failing.

BudgieInWA:
Thanks for merging the module restructure @droogmic, I have just merged the newer changes, so this should merge cleanly (currently :P).

I don't know anything about the kotlin tests, and the report linked here didn't show me what was wrong, so can someone else please look into that? Feel free to push commits directly to this branch (which I think is enabled by the "Allow edits by maintainers" checkbox that I have on).

I would like for this PR to be merged now, if we are happy with the direction it takes us. I am happy to be assigned followup issues (or be pinged in fixes to nitpicky issues so I can learn from them) in the interest of working from the same branch going forward. In future, my PRs will be more focused than this one: The detours and explorations that I did on this PR were valuable to me in learning rust and learning this codebase -- especially with your input -- but that's not the way I like to do PRs in general; and I don't plan on restructuring things out of the blue as I ended up doing in this PR.

rust/osm2lanes/src/transform/tags_to_lanes/lane.rs
```diff
@@ -0,0 +1,276 @@
...
+                            bothways,
+                        }
+                    }
+                },
+            }
+        }
+    }
+}
+
+// struct TurnLanesScheme {
```
dabreegster, line 227: nitpick: Remove large chunks of commented code, or use `#[allow(unused)]`. `TurnDirections` sounds like it should wind up being an enum anyway.

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -615,15 +661,45 @@ pub fn tags_to_lanes(
...
+            .map_or(0, |o| if o.is_some() { 1 } else { 0 }),
+        backward: busway
+            .backward_side_direction
+            .map_or(0, |o| if o.is_some() { 1 } else { 0 }),
+        bothways: Infer::Default(0),
+    };
+    let centre_turn_lanes = CentreTurnLaneScheme::new(tags, oneway, locale, &mut warnings);
+    let lanes = LanesScheme::new(
+        tags,
+        /*highway,*/ oneway,
```
dabreegster, line 685: Same here -- bits of commented code checked in can add up quickly and get overwhelming. (I'm going to pick on another project I've worked on, https://github.com/Urban-Analytics/RAMP-UA/blob/c5ecee2ab8eebd2b9cffc41956c7c83195b709d1/coding/initialise/population_initialisation.py#L386)

dabreegster (approved):
I'm strongly in favor of merging and iterating with smaller cleanups, instead of having more divergence with #138 build up. No worries on a large PR or roundabout way of getting to this point -- learning Rust is (at least) half the fun! Some pretty neat ideas came out of this too.

I'll glance at the kotlin problem -- I spotted something about JSON parsing in the logs.

dabreegster:
Kotlin: I can't follow https://github.com/a-b-street/osm2lanes/blob/main/CONTRIBUTING.md#kotlin locally; gradle is throwing some unhelpful error. The test log shows it succeeded at YAML -> JSON, but then `kotlinx.serialization.json.internal.JsonDecodingException at Test.kt:56`.  Unhelpfully, https://github.com/a-b-street/osm2lanes/blob/main/kotlin/src/test/kotlin/Test.kt doesn't have 56 lines, so it's maybe some test framework file?

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
BudgieInWA: Sensible, I'll do a pass, and get some TODOs too.

data/tests.yml
```diff
@@ -846,9 +846,10 @@
   skip_kotlin: true
   tags:
     highway: "road"
-    lanes: "3"
     sidewalk: "no"
     shoulder: "no"
+    lanes: "3"
+    lanes:forward: "2"
```
droogmic, line 852: can this not be inferred from `bus:lanes:forward`
what happens when this is removed?

rust/osm2lanes/src/locale.rs
```diff
@@ -171,6 +171,15 @@ impl std::str::FromStr for DrivingSide {
     }
 }
 
+impl ToString for DrivingSide {
+    fn to_string(&self) -> String {
+        match self {
+            Self::Right => String::from("right"),
```
droogmic, line 177: not sure if important,

but I avoided implementing this to avoid confusing right hand drive with right hand travel.

data/tests.yml
droogmic, line 852: can this not be inferred from bus:lanes:forward
what happens when this is removed?

rust/osm2lanes/src/locale.rs
droogmic, line 177: not sure if important,

but I avoided implementing this to avoid confusing right hand drive with right hand travel.

rust/osm2lanes/src/tag/mod.rs
```diff
@@ -11,6 +11,8 @@ pub use key::TagKey;
 mod osm;
 pub use osm::{Highway, HighwayType, Lifecycle, HIGHWAY, LIFECYCLE};
 
+use crate::transform::{RoadWarnings, TagsToLanesMsg};
```
droogmic, line 14: it feels strange including these in this mod.

I would expect tag to eventually become its own crate or similar, with no dependencies on the osm2lanes transformation logic.

rust/osm2lanes/src/tag/mod.rs
```diff
@@ -90,6 +92,26 @@ impl Tags {
         self.map.get(k.as_ref()).map(String::as_str)
     }
 
+    /// Get the value for the given key and parse it into T.
+    /// Add a `TagsToLanesMsg::Unsupported` if parsing fails.
+    pub fn get_parsed<K: AsRef<str>, T: FromStr>(
+        &self,
+        key: &K,
+        warnings: &mut RoadWarnings,
+    ) -> Option<T> {
```
droogmic, line 101: `Result<Option<T>, String>`

where the Err is the value that cannot be parsed?

or

`Result<Option<T>, E>`

where E comes from the underlying FromStr

droogmic: maybe Option<Result<T>> makes more sense?

rust/osm2lanes/src/transform/tags_to_lanes/lane.rs
```diff
@@ -0,0 +1,266 @@
+use super::{Infer, Oneway};
+use crate::locale::Locale;
+use crate::road::Direction;
+use crate::tag::{TagKey, Tags};
+use crate::transform::tags_to_lanes::modes::{BuswayScheme, LanesBusScheme};
+use crate::transform::{RoadWarnings, TagsToLanesMsg};
+
+const LANES: TagKey = TagKey::from("lanes");
+#[derive(Debug)]
+pub struct LanesScheme {
```
droogmic, line 10: the name of the mod and this struct are unclear for me

in my PR I named this `Counts`, as it isn't really defining a scheme?

rust/osm2lanes/src/transform/tags_to_lanes/lane.rs
```diff
@@ -0,0 +1,266 @@
...
+    pub lanes: Infer<usize>,
+    pub forward: Infer<usize>,
+    pub backward: Infer<usize>,
+    pub bothways: Infer<usize>,
```
droogmic, line 14: `both_ways`.
Rust and the OSM tag both use snake case.

rust/osm2lanes/src/transform/tags_to_lanes/lane.rs
```diff
@@ -0,0 +1,266 @@
...
+                    }
+                },
+            }
+        }
+    }
+}
+
+const CENTRE_TURN_LANE: TagKey = TagKey::from("centre_turn_lane");
+pub struct CentreTurnLaneScheme {
+    pub present: Infer<bool>,
```
droogmic, line 232: what does "present" mean in this context?

rust/osm2lanes/src/transform/tags_to_lanes/modes/bus.rs
```diff
@@ -23,6 +23,21 @@ impl LaneBuilder {
     }
 }
 
+const BUSWAY: TagKey = TagKey::from("busway");
+
+#[derive(Debug)]
+pub struct BuswayScheme {
+    pub forward_side_direction: Infer<Option<Direction>>,
+    pub backward_side_direction: Infer<Option<Direction>>,
+}
```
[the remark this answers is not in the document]
droogmic: second: this would be nice to add this as a standalone intermediate refactor PR, to make reviewing easier.

droogmic (changes requested):
I can agree to this once you look at what I have done and can justify why this is better.

For example:
How do we extend this to include <https://taginfo.openstreetmap.org/keys/taxi:lanes>?
it feels a lot of the lane handling is hardcoded to bus and only bus?

rust/osm2lanes/src/transform/tags_to_lanes/modes/bus.rs
```diff
@@ -225,7 +276,7 @@ fn bus_lanes(
             }
             if let Some(backward) = backward {
                 let backward_access = Access::split(backward).map_err(|a| {
-                    RoadError::from(TagsToLanesMsg::unsupported(
+                    RoadError::Msg(TagsToLanesMsg::unsupported(
```
droogmic, line 279: do you remember why you needed to change this? it works as well, I just don't know which is better :P

rust/osm2lanes/src/transform/tags_to_lanes/modes/mod.rs
```diff
@@ -4,8 +4,7 @@ mod bicycle;
 pub(super) use bicycle::bicycle;
 
 mod bus;
-pub(super) use bus::bus;
-
+pub(super) use bus::{bus, BuswayScheme, LanesBusScheme};
```
droogmic, line 7: caution, whitespace inconsistency

rust/osm2lanes/src/tag/mod.rs
BudgieInWA: Agreed, `get_parsed` is a convenience function for parsing a bunch of tags at the same time, and should probably live in `transform` somewhere. It was suggested that it be moved in here, but maybe it should be a function, not a method: `let l: Option<usize> = get_parsed_tag(tags, "lanes", &warnings)`; or a method on `RoadWarnings`: `let l: Option<usize> = warnings.parse_tag(tags, "lanes")`?

rust/osm2lanes/src/tag/mod.rs
BudgieInWA: It would be `Result<T, E>` (which would be converted to an `Option` with `ok()` most of the time, because the error is already added to the `warnings`).

rust/osm2lanes/src/transform/tags_to_lanes/modes/bus.rs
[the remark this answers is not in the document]
BudgieInWA: > I don't think I found a case of a Backward in forward_side_direction or Forward in backward_side_direction
nor would I expect it.

Given this, I agree that `BuswayScheme(Infer<(bool, bool)>)` is better.

rust/osm2lanes/src/transform/tags_to_lanes/lane.rs
BudgieInWA: I didn't know what to name the mod. The struct is named `LanesScheme` because it represents the `lanes:*` tags (which together make up what I was calling "scheme").

BudgieInWA:
> I can agree to this once you look at what I have done and can justify why this is better.
>
> How do we extend this to include [taginfo.openstreetmap.org/keys/taxi:lanes](https://taginfo.openstreetmap.org/keys/taxi:lanes)?
> it feels a lot of the lane handling is hardcoded to bus and only bus?

My intention is to achieve a couple of things:

1. Separate tag parsing from road building, by introducing `Scheme` structs that represent the valid states described by smalls sets of tags. `BuswayScheme` is a good example: the `busway` tags are a complicated way of describing "is there a bus lane on the forward side, and is there a bus lane on the backward side", so we parse all those tags into `BuswayScheme(Infer<(bool, bool)>)` and never have to look up a `busway` tag value again.
2. Fix the guessing of the total number of lanes, by inverting the order that the tags are looked at.
3. Move towards tackling the ordering of looking at different tags at the top level `tags_to_lanes`. a.k.a. express the direction of dependencies between tags in one place. I have only taken a step in that direction, creating `BuswayScheme`, `CentreTurnLaneScheme`, and `LanesScheme` (which take other `Scheme` arguments to show dependencies), and constructing them in the right order at the top level.

These ideas would be further helped by:

- Implementing a `Scheme` for every tag or set of tags that we parse (even `Oneway` has way more complexity that we don't support yet).
- Bringing the guessing logic out of the `Scheme` constructors, and into road builder. Maybe `Scheme` constructors should not `Infer::Guess` at all, and only return `Direct`, `Calculated` (for internally calculated values, like `lanes=lanes:forward + lanes:backward`) and `Default` values. At the top level, or as a set of `RoadBuilder` methods, follow an ordered procedure for building up the road using the `Scheme`s, but no tags.
   - This would allow us to remove all dependencies from `LanesScheme::new`, (except maybe `oneway`) and figure out the guesses in a context which has access to a `RoadBuilder` with some lanes already added and the parsed `LanesScheme` (like https://github.com/a-b-street/osm2lanes/issues/136#issuecomment-1104149579). That step would treat `Direct` values in `LanesScheme` differently to `Default` values.

droogmic:
Ok, but then I think we might be trying to do too much in one PR (which is where the difficulty is coming from).

Either we fix the lane counting, or we fix the scheme parsing. Doing both at the same time is blocking quick progress :)

rust/osm2lanes/src/locale.rs
BudgieInWA: I am confused about the difference, so maybe it is important! Is right hand drive when the driver gets in the right hand door of the car? I suppose this should be `Into<TagKey>` instead of `ToString` or something?

data/tests.yml
BudgieInWA: It could, but `lanes:forward` should be tagged in OSM anyway. Inferring it is being a bit fancy and I'm not trying to solve the entirety of lane tagging in this PR.

rust/osm2lanes/src/transform/tags_to_lanes/lane.rs
BudgieInWA: It refers to the presence of such a lane: "is a centre turn lane present on this road?".

Is something like `struct CentreTurnLaneScheme(Infer<bool>)` possible? I haven't used that language feature before.

rust/osm2lanes/src/transform/tags_to_lanes/modes/bus.rs
BudgieInWA: Something silly from my IDE while I was merging. Turns out that wasn't what was breaking the build. We can probably roll this back and I can turn off that warning somehow.
from error

data/tests.yml
droogmic: > but lanes:forward should be tagged in OSM anyway

it _should_ but it often isn't, we shouldn't depend on perfect OSM tagging.
we can leave it in to simplify the implementation for now, but please add a comment describing why this data is redundant so that we can try to remove it later?

rust/osm2lanes/src/transform/tags_to_lanes/modes/bus.rs
droogmic: it's fine, I was just curious

rust/osm2lanes/src/transform/tags_to_lanes/lane.rs
droogmic: yes, that would be the correct construct.

droogmic: https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-the-newtype-pattern-to-implement-external-traits-on-external-types

droogmic:
kotlin is probably failing in the tests.yml file, so I suggest bisecting the changes you made there until kotlin is happy?

droogmic:
my bad, the button for close is very close to the one with comment :laughing:

data/tests.yml
```diff
@@ -1158,68 +1161,105 @@
...
+        designated: motor_vehicle
+      - type: shoulder
+
+- description: assume lanes:forward to be lanes - lanes:backward
+  skip_python: true
+  skip_kotlin: true
+  tags:
+    highway: secondary
+    lanes: 4
+    lanes:backward: 1
```
droogmic, line 1226: these are interpreted as integers, when they should be strings, that causes the kotlin tests to fail

dabreegster:
Thanks for splitting into smaller PRs and pushing them through. Just checking in, was closing this PR intentional, or a side effect of the commit message there? Are any of the ideas in this PR still not merged in by one of the smaller PRs?

droogmic:
A side-effect. I removed a lot of stuff from this PR that were not needed to make the tests pass

@BudgieInWA will probably want to go through the final status to reintroduce some of the less critical ideas he had (I removed the Infer::Guessed variant for example), preferably as a TDD when they become needed.

BudgieInWA:
> A side-effect. I removed a lot of stuff from this PR that were not needed to make the tests pass
> 
> @BudgieInWA will probably want to go through the final status to reintroduce some of the less critical ideas he had (I removed the Infer::Guessed variant for example), preferably as a TDD when they become needed.

This is a great outcome. Getting these ideas our into the open is probably the appropriate amount of progress for now.

PR #112, opened 2022-03-26
Refactor Serde Spec and Tests

droogmic:
To give us a chance to slowly iterate on refactoring the tests :)

### DIFF
--- a/data/tests.yml
+++ b/data/tests.yml
@@ -59,31 +59,33 @@
     ignore_warnings: true
   skip_kotlin: true
   tags:
-    highway: trunk
-    surface: asphalt
+    highway: "trunk"
+    surface: "asphalt"
   driving_side: left
   ISO 3166-2: AU-WA
-  output:
-  - type: shoulder
-  - type: separator
-    markings:
-    - style: solid_line
-      color: white
-  - type: travel
-    direction: forward
-    designated: motor_vehicle
-  - type: separator
-    markings:
-    - style: dotted_line
-      color: white
-  - type: travel
-    direction: backward
-    designated: motor_vehicle
-  - type: separator
-    markings:
-    - style: solid_line
-      color: white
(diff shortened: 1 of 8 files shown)


rust/osm2lanes/src/road/lane.rs
```diff
@@ -5,26 +5,31 @@ use crate::{Locale, Metre, Speed};
 
 /// A single lane
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-#[serde(tag = "type")]
+#[serde(tag = "type", rename_all = "lowercase")]
```
BudgieInWA, line 8:
```suggestion
#[serde(tag = "type", rename_all = "snake_case")]
```

BudgieInWA (approved):
Seems like a step forward to me. :+1:

data/tests.yml
```diff
@@ -59,31 +59,33 @@
...
+    - type: travel
+      direction: backward
+      designated: motor_vehicle
+    - type: separator
+      markings:
+      - style: solid_line
+        color: white
+    - type: shoulder
 
 - description: "Alaska Highway, a 2 lane rural road."
```
dabreegster, line 90: Is it easy to hack up a Python script to change all the rest automatically and maybe have to deal with the few places where we actually need to specify the `highway` in the output? (Or if it's all of them, copy it over from the input?)

dabreegster (approved):

data/tests.yml
droogmic: I find it fun always trying to do incremental changes, because the moment you cannot do it in one go, you really need those skills :P

dabreegster: I'll for a steady push forward, this comment wasn't a blocker! Could even be a good first issue if someone else wants to poke at the project

PR #113, opened 2022-03-26
Rust: Lane Separator Colour by Locale

droogmic:
We now have a problem, 2 crates about countries...

I think locale-codes does more of what we want, but it also isn't perfect :(.


BudgieInWA:
I think this might be a good time to rename `rust.ignore_warnings` to `expect_warnings` or something similar in `tests.yml`.

I don't have an opinion in the country crate.

rust/osm2lanes/src/test.rs
```diff
@@ -377,6 +377,10 @@ mod tests {
                 let expected_road = test.expected_road();
                 match road_from_tags {
                     Ok(road_from_tags) => {
+                        if test.test_has_warnings() && road_from_tags.warnings.is_empty() {
+                            test.print();
+                            println!("Expected warnings. Try removing `ignore_warnings`.");
```
dabreegster, line 382: Should this be a panic / error? If the test passes in Rust, STDOUT isn't shown at all

dabreegster, line 382: Should this be a panic / error? If the test passes in Rust, STDOUT isn't shown at all

dabreegster (approved):

PR #114, opened 2022-03-26
Rust: Enable Benchmarking

droogmic:
If we can configure the settings correctly, a bot should comment here with a nice looking table.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -105,10 +105,10 @@ jobs:
       uses: Swatinem/rust-cache@v1
       with:
         working-directory: rust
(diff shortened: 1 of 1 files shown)


droogmic:
@dabreegster, I don't think I have permissions to change this repo's settings?

dabreegster:
I set role = admin, let me know if that works

droogmic:
> I set role = admin, let me know if that works

Ah, I see now that I am running into this.

If I have more time to play with this I will try to do something for fun in 2 steps.

droogmic:
For now, if you want to check benchmarks from a fork, you need to open up the logging.

PR #115, opened 2022-03-26
Rust Update Toolchain and Clippy Pedantic

droogmic:
Update the nightly toolchain, and try to use clippy pedantic. It is _very_ pedantic.


dabreegster (approved):
I skimmed through, everything looks reasonable to me. Stricter compile-time rules always nice to see

PR #118, opened 2022-04-07
Rust: Increment Tests and Separator

droogmic:
YAML Reformat
Fix Highway Inner Edge

### DIFF
--- a/data/tests.yml
+++ b/data/tests.yml
@@ -4,8 +4,8 @@
 ### - simple to complex
 ### - rural to urban
 ### - By infrastructure:
-###   - car-centric 
-###   - pedestrian 
+###   - car-centric
+###   - pedestrian
 ###   - cycleways
 ###   - busways
 ###   - heavy goods vehicles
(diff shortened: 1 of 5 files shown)


data/tests.yml
```diff
@@ -172,70 +176,75 @@
   mapillary: https://www.mapillary.com/app/?pKey=2655760771400441
   skip_python: true
   skip_kotlin: true
-  rust: false
+  rust:
+    expect_warnings: true
```
droogmic, line 180: note, this test was fixed :D 

one at a time ;)

data/tests.yml
```diff
@@ -4,8 +4,8 @@
 ### - simple to complex
 ### - rural to urban
 ### - By infrastructure:
-###   - car-centric
```
dabreegster, line 7: We've hit "large diff" territory -- I'm trusting changes here blindly!

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -179,6 +180,12 @@ impl LaneBuilder {
             None => panic!(),
         }
     }
+
+    /// Create a mirrored version of the lane
+    #[must_use]
+    fn mirror(&self) -> &Self {
+        self
```
dabreegster, line 187: Shouldn't this call something?

dabreegster (approved):

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
droogmic: I wanted to get the interface in. For now it actually doesn't need to do anything, because none of the lanes are handed (no chirality)

droogmic:
```suggestion
        # TODO: this doesn't need to do anything for now
        # check back after v1.0.0 to see if this is still the case
        self
```

data/tests.yml
droogmic: yup, I highlighted with my comment the last meaningful change

PR #119, opened 2022-04-08
Rust: refactor tags_to_lanes into modules

droogmic:
There are no meaningful changes.

Sorry for the merge conflicts it may cause.


dabreegster (approved):
Do you expect more modules in `tags_to_lanes` that don't count as modes? Otherwise the extra nesting doesn't seem necessarily helpful

droogmic:
> Do you expect more modules in `tags_to_lanes` that don't count as modes? Otherwise the extra nesting doesn't seem necessarily helpful

I expect an extensive tree in modes/* and separator/*, and maybe a few flat files in between.

dabreegster:
Seems good to me, then!

PR #120, opened 2022-04-08
Rust: Add Support for Path and Steps

droogmic:
Small increment to support way 40297361

### DIFF
--- a/rust/osm2lanes/src/transform/lanes_to_tags/mod.rs
+++ b/rust/osm2lanes/src/transform/lanes_to_tags/mod.rs
@@ -56,6 +56,12 @@ impl Lane {
 pub fn lanes_to_tags(lanes: &[Lane], locale: &Locale, config: &Config) -> TagsResult {
     let mut tags = Tags::default();
     let mut oneway = false;
+
+    if !lanes.iter().any(|lane| lane.is_motor() || lane.is_bus()) {
+        tags.checked_insert("highway", "path")?;
+        return Ok(tags);
+    }
+
     tags.checked_insert("highway", "road")?; // TODO, add `highway` to `Lanes`
     {
         let lane_count = lanes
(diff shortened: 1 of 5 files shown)


rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -449,9 +449,6 @@ impl RoadBuilder {
                 }
                 _ => todo!(),
             };
-            let forward_edge = dbg!(forward_edge);
-            let middle_separator = dbg!(middle_separator);
-            let backward_edge = dbg!(backward_edge);
```
droogmic, line 454: I need to find a way to stop this being committed in the future.

rust/osm2lanes/src/transform/lanes_to_tags/mod.rs
```diff
@@ -56,6 +56,12 @@ impl Lane {
 pub fn lanes_to_tags(lanes: &[Lane], locale: &Locale, config: &Config) -> TagsResult {
     let mut tags = Tags::default();
     let mut oneway = false;
+
+    if !lanes.iter().any(|lane| lane.is_motor() || lane.is_bus()) {
+        tags.checked_insert("highway", "path")?;
```
dabreegster, line 61: What if this method took `&Road` and we just used `Highway` here? This seems super brittle; what if it was `steps` or some other kind of footway?

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
dabreegster: Ah, I noticed in code review but didn't comment, my mistake. If it was a big enough problem, we could have a presubmit check grep for it or something

rust/osm2lanes/src/transform/lanes_to_tags/mod.rs
droogmic: yeah, this will need to change, will refactor as part of a future PR :)

dabreegster (approved):

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
BudgieInWA: That sounds like the way to go! I noticed [Jon Blow](https://www.youtube.com/user/jblow888) using `// @nocheckin` (and related) in his setup and thought it was a useful idea.

Issue #100, opened 2022-03-19
Represent dividers semantically


BudgieInWA:
I want to propose that dividers be represented internally by their semantics, instead of their visual/physical manifestation, and be translated into specific markings in a separate step. My initial feeling is that it would shift most or all locale dependent questions about dividers out of the tag parsing step, into its own *divider to markings* step. Providing the divider semantics in the output would also be useful for users.

## Semantic divider types

Dividers would be broadly categorised something like:

- curb
- road edge line (inner / outer)
- lane separating line
- center line
- median (see below)

with additional semantic properties such as

- stopping: default/no_standing/no_stopping/parking?/...
- can_change: no/only_left/only_right/both ([change=](https://wiki.openstreetmap.org/wiki/Key:change) and [overtaking=](https://wiki.openstreetmap.org/wiki/Key:overtaking)

## `dividers_to_markings`

A separate function concerned with just the problem of converting semantic dividers to markings in the context of a given road would be a easier for new contributors to contribute to, which would be valuable for capturing global nuances. (I would be interested to see how static the mappings turn out to be.)

Having the markings be a separate step would also be a useful to allow additional information to be added to the road after examing the semantics. E.g. I have two connected OSM ways: 2 lanes into 3 lanes. I get the semantic lane info, compare them and decide that first lane in the 3 lane way is an added lane. I annotate the appropriate divider as "merging" before getting markings, and receive a beautifully appropriate dotted line (here in Oz).

Exposing the function as part of the API would also be a good way to query the lane marking database, because describing a situation from scratch in the output schema is much easier than describing it in the OSM schema!

## Divider widths

Given that different locales might have different width markings for the same semantic divider, semantic dividers would have zero width: their markings fit within the the width(s) of the adjacent lane(s). (I'm not even sure this isn't how it already works). This is how I would prefer to receive osm2lanes output in semantic mode as a user.

## Medians

This makes the "median" divider the odd one out, as the only divider with width of its own. Perhaps this suggest that the median itself should be a "lane" like a shoulder or a verge. One benefit is that the markings that divide the road from the median can be described too. In Australia, a median will often have a continuous solid line surrounding it, even as it transitions from a solid colour to a raised median to a diagonal hatching. The following output to describe the median would make sense to me:

![A continuous line (divider) surrounds the median (buffer)](https://user-images.githubusercontent.com/3784591/159111353-bfa7e58e-0b33-479e-b81f-ea8d2fc83f68.png)

```js
[...]

PR #125, opened 2022-04-10
Rust: Add Semantic Separator Example

droogmic:
This related to issue #100

### DIFF
--- a/rust/osm2lanes/src/transform/tags_to_lanes/separator/semantic.rs
+++ b/rust/osm2lanes/src/transform/tags_to_lanes/separator/semantic.rs
@@ -1,8 +1,9 @@
-use crate::metric::Metre;
-use crate::road::Style;
+use crate::metric::{Metre, Speed};
+use crate::road::{Designated, Style};
+use crate::transform::tags_to_lanes::Infer;
 
 /// Semantic speed class
-enum _SpeedClass {
+pub enum SpeedClass {
     Walking,
     /// ~30kph / ~20mph
     Living,
@@ -13,18 +14,59 @@ enum _SpeedClass {
     Max,
 }
 
+impl From<Speed> for SpeedClass {
+    fn from(s: Speed) -> Self {
+        match s.kph() {
+            s if s > 0.0_f64 && s <= 10.0_f64 => Self::Walking,
+            s if s > 10.0_f64 && s < 40.0_f64 => Self::Living,
+            s if s >= 40.0_f64 && s < 70.0_f64 => Self::Intra,
+            s if s >= 70.0_f64 && s < 100.0_f64 => Self::Inter,
+            s if s >= 100.0_f64 && s < 200.0_f64 => Self::Max,
+            _ => panic!("unexpected speed {s}"),
+        }
+    }
+}
+
+/// Overtaking rules
+///
+/// Note: this does not take into account the local vehicle-specific rules,
+/// e.g. for motorcycle filtering or overtaking tractors.
+pub enum Overtaking {
+    Permitted,
+    Prohibited,
+}
+
+impl Default for Overtaking {
+    fn default() -> Self {
+        // fail-deadly, see README
+        Self::Permitted
+    }
+}
+
 /// Semantic lane separator
-enum _Separator {
+pub enum Separator {
     /// Into grass or dirt
     SoftEdge,
     /// Into a building or other hard surface
     HardEdge,
     /// Motorway (or other) shoulder
-    Shoulder(_SpeedClass),
-    /// Road paint for same direction
-    Lane { speed: _SpeedClass },
-    /// Road paint for opposite direction
-    Centre { speed: _SpeedClass },
+    Shoulder { speed: Infer<SpeedClass> },
+    /// Road paint between same direction
+    Lane {
+        speed: Infer<SpeedClass>,
+        overtaking: Overtaking,
+    },
+    /// Road paint between opposite direction
+    Centre {
+        speed: Infer<SpeedClass>,
+        overtaking: Overtaking,
+    },
+    /// Road paint between different modes
+    // TODO: solve directionality
+    Modal {
+        inside: Designated,
+        outside: Designated,
+    },
     /// Painted area
     Buffer { width: Metre, style: Style },
     /// Kerb step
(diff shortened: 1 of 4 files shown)


rust/osm2lanes/src/transform/tags_to_lanes/separator/semantic.rs
```diff
@@ -0,0 +1,35 @@
...
+/// Semantic lane separator
+enum _Separator {
+    /// Into grass or dirt
+    SoftEdge,
+    /// Into a building or other hard surface
+    HardEdge,
+    /// Motorway (or other) shoulder
+    Shoulder(_SpeedClass),
+    /// Road paint for same direction
+    Lane { speed: _SpeedClass },
```
dabreegster, line 25: Is speed class important here? For lane and centre separators, I'd imagine the "can you change lanes here" and "can you over-take (against opposite-flowing traffic)" are the attributes we may want to refine

droogmic: I think in some countries the paint style differs based on speed class, the classic example is multiple lanes on a motorway vs multiple lanes in an inner-city artery?

But agreed, I am missing the overtake part :)

dabreegster: Having read #100 more carefully, I understand where this is going and am in favor. :)

rust/osm2lanes/src/transform/tags_to_lanes/separator/semantic.rs
```diff
@@ -0,0 +1,79 @@
...
+        speed: Infer<SpeedClass>,
+        overtaking: Overtaking,
+        more_than_2_lanes: bool,
+    },
+    /// Road paint between different modes
+    // TODO: solve directionality
+    Modal {
+        inside: Designated,
+        outside: Designated,
+    },
```
BudgieInWA, line 71: I think only details that are properties of the separator should be part of this type, but the `separator to markings` transformations should have access to both adjacent lanes and the road. The speed can be different between lanes, but can that effect the marking style per lane? Also, I am sure that different locales use all sorts of exotic properties to determine marking styles. 

Then `Modal` could be removed in favour of `Lane` (possibly with `different_modes: bool` for convenience) and the directionality question could be dodged.

rust/osm2lanes/src/transform/tags_to_lanes/separator/semantic.rs
```diff
@@ -0,0 +1,79 @@
...
+    _HardEdge,
+    /// Motorway (or other) shoulder
+    Shoulder { speed: Infer<SpeedClass> },
+    /// Road paint between same direction
+    Lane {
+        speed: Infer<SpeedClass>,
+        overtaking: Overtaking,
+    },
+    /// Road paint between opposite direction
+    Centre {
```
BudgieInWA, line 61: Maybe the variants that are just paint should have "Marking" in their names? Like `CenterMarking`, `LaneMarking`, `MarkedBuffer`, and `ShoulderMarking` (or `EdgeMarking`), to avoid confusion with other types and remind us that physical separators will use a different variant.

rust/osm2lanes/src/transform/tags_to_lanes/separator/semantic.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// Into grass or dirt
+    _SoftEdge,
+    /// Into a building or other hard surface
+    _HardEdge,
+    /// Motorway (or other) shoulder
+    Shoulder { speed: Infer<SpeedClass> },
+    /// Road paint between same direction
+    Lane {
+        speed: Infer<SpeedClass>,
+        overtaking: Overtaking,
```
BudgieInWA, line 58: OSM uses the `change` tag between any two lanes, and `overtaking` exclusively for crossing the centreline, which is how I understand the term colloquially. I think this field should be `changing: bool` or `changing: LaneChanging`. (In AU, we have dashed lines when changing is allowed, but dotted lines when a lane is beginning, but I don't think the distinction can be tagged...)

BudgieInWA (approved):
I like the direction this is going!

How to describe a median is still a tricky one, but working with this system  for some time might shed some light on it.

rust/osm2lanes/src/transform/tags_to_lanes/separator/semantic.rs
droogmic: >  only details that are properties of the separator should be part of this type

What is the semantic difference then between `Centre` vs `Lane` then?

to me there is a semantic difference between markings separating direction, markings separating lanes of the same type and direction, and markings separating modes?

rust/osm2lanes/src/transform/tags_to_lanes/separator/semantic.rs
droogmic: I could also nest the enum:

```
    ...
    Marking(Marking),
    ...

    enum Marking {
        Centre,
        Lane,
    }
```

rust/osm2lanes/src/transform/tags_to_lanes/separator/semantic.rs
BudgieInWA: I see your point. A line separating a car lane from a bike lane *does* feel very different from a line separating two car lanes,  but a line separating a car lane from a bus lane is much more similar to the latter -- especially because bus lanes can often be used by cars to make turns or during part of the day. As long as `Centre` takes precedence,  `Modal` seems like a perfectly sensible distinction.

rust/osm2lanes/src/transform/tags_to_lanes/separator/semantic.rs
BudgieInWA: Hmm.. I think I prefer a flat enum. `Separator::Centre` is also clear, but my novice ruse instinct was to `use Separator::*` in my head :P I would not object to whichever you choose.

rust/osm2lanes/src/transform/tags_to_lanes/separator/semantic.rs
droogmic: > especially because bus lanes can often be used by cars to make turns or during part of the day

dabreegster will know that bus lanes in london are often shared between cyclists and busses and taxis.

Where I live all 3 modes are almost entirely always separated where possible.

I suggest sticking to our use of `designated` to mean the primary mode, and separators between those consisting only of paint should for a semantic category.

PR #128, opened 2022-04-11
Rust: Refactor Web into Components

droogmic:
No functional changes, preparation work for other work in improving web view.

### DIFF
--- a/rust/osm2lanes-web/src/main.rs
+++ b/rust/osm2lanes-web/src/main.rs
@@ -1,19 +1,25 @@
+use std::cell::RefCell;
+use std::rc::Rc;
 use std::str::FromStr;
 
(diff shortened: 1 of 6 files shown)


droogmic:
(no detailed review expected, like usual for a pure refactor...)

rust/osm2lanes-web/src/main.rs
```diff
@@ -74,8 +55,7 @@ pub enum Msg {
 }
 
 pub struct App {
-    focus_ref: NodeRef,
-    state: State,
+    state: Rc<RefCell<State>>,
```
dabreegster, line 58: Where was the need for these two introduced? Is it because some of the methods got moved to other modules, and the borrow checker thinks there's a double borrow of two different pieces of `App` at the same time?

dabreegster (approved):

Issue #132, opened 2022-04-14
Living streets


dabreegster:
`cargo run way 1005922755` yields `"error": "unimplemented: highway=living_street"` (https://www.openstreetmap.org/way/1005922755)

Few thoughts:

1) With all the indirections around error transformations, it's not easy to spot the place in the source that's actually deciding this tag isn't implemented. I searched for `unimplemented_tag`, and the 3 places don't seem relevant. Should we consider one of the error handling crates that adds a bit of context of file + line number of the "thrown" error?

2) I'd add a test case, except I don't know what we want to infer about `highway=living_street` (the `access=destination` doesn't seem relevant). Standard 1 travel lane and sidewalk each side?

BudgieInWA:
A default living street sounds like 1 wide traffic lane in each direction, designated foot and vehicle, speed limit "walking speed" depending on locale.

droogmic:
> Should we consider one of the error handling crates that adds a bit of context of file + line number of the "thrown" error

I looked into that last week, and found that we apparently should use https://docs.rs/thiserror/latest/thiserror/, except it doesn't do the line magic stuff for us, so we need to use https://doc.rust-lang.org/stable/std/panic/struct.Location.html ourselves maybe.

droogmic:
https://wiki.openstreetmap.org/wiki/Tag:highway%3Dliving_street
after reading this and correlating it against experiences in various countries, I think the default (no locale) should be:
a single lane of travel, no direction, and no designated traffic, max_speed defaults to 20, e.g.
```
[{
    type: travel
    designated: Any
    max_speed: 20,
}]
```
This requires reviving https://github.com/a-b-street/osm2lanes/blob/main/rust/osm2lanes/src/road/lane.rs#L79

BudgieInWA:
This reminds me of the question on the OSM side about tagging "2 lane" roads without lane markings: I don't remember where I read it, but it seems like the a road with no markings that is wide enough for 2 cars to pass each other would be tagged `lanes=2` because traffic uses the road as if it has one lane in each direction. So, if vehicles on a living street are expected to keep to their side etc. then I think 2 lanes would be appropriate. Using a single lane has its own benefits too, so I'm not against it.

droogmic:
@BudgieInWA We cover this in our readme :)
https://github.com/a-b-street/osm2lanes#examples-4

westnordost:
> speed limit

[...]

PR #133, opened 2022-04-16
Rust: Add Error Locations

droogmic:
Use std::panic::Location to idenitfy error locations
Refactoring with thiserror, another crate, or without may cut down on boilerplate


droogmic:
an ugly, PR, I could try and split it in one or two

dabreegster (approved):
Things moving to a new module at the same time as adding `location` was a little confusing, but no need to split -- everything looks good, and `"error": "unimplemented: 'highway=living_street' - osm2lanes/src/transform/tags_to_lanes/mod.rs:273:32"` nicely solves #132 :)   Thanks!

PR #134, opened 2022-04-18
Rust Lane Count

droogmic:
An alternative to https://github.com/a-b-street/osm2lanes/pull/111

I copied over a part of @BudgieInWA, complementing it with a hybrid way of handling bus ways...

It is a bit messy, so I can start to split into multiple PRs if needed.

### DIFF
--- a/rust/osm2lanes/src/tag/mod.rs
+++ b/rust/osm2lanes/src/tag/mod.rs
@@ -11,7 +11,7 @@ pub use key::TagKey;
 mod osm;
 pub use osm::{Highway, HighwayType, Lifecycle, HIGHWAY, LIFECYCLE};
 
-use crate::transform::{RoadMsg, RoadWarnings};
+use crate::transform::{RoadWarnings, TagsToLanesMsg};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct DuplicateKeyError(String);
@@ -92,8 +92,9 @@ impl Tags {
         self.map.get(k.as_ref()).map(String::as_str)
     }
 
-    /// Get the value for the given key and parse it into T. Add a RoadMsg::Unsupported if parsing
-    /// fails.
+    /// Get the value for the given key and parse it into T.
+    /// Add a RoadMsg::Unsupported if parsing fails.
+    // TODO: check if interface should be moved to transform
     pub fn get_parsed<K: AsRef<str>, T: FromStr>(
         &self,
         key: &K,
@@ -102,7 +103,10 @@ impl Tags {
         self.get(key).and_then(|val| match val.parse::<T>() {
             Ok(n) => Some(n),
             Err(_) => {
-                warnings.push(RoadMsg::unsupported_tag(key.as_ref().to_owned(), val));
+                warnings.push(TagsToLanesMsg::unsupported_tag(
+                    key.as_ref().to_owned(),
+                    val,
+                ));
                 None
             },
         })
--- a/rust/osm2lanes/src/tag/mod.rs
+++ b/rust/osm2lanes/src/tag/mod.rs
@@ -11,6 +11,8 @@ pub use key::TagKey;
 mod osm;
 pub use osm::{Highway, HighwayType, Lifecycle, HIGHWAY, LIFECYCLE};
 
+use crate::transform::{RoadMsg, RoadWarnings};
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct DuplicateKeyError(String);
 
@@ -88,6 +90,22 @@ impl Tags {
     /// Get value from tags given a key
     pub fn get<T: AsRef<str>>(&self, k: T) -> Option<&str> {
         self.map.get(k.as_ref()).map(String::as_str)
+    }
+
+    /// Get the value for the given key and parse it into T. Add a RoadMsg::Unsupported if parsing
+    /// fails.
+    pub fn get_parsed<K: AsRef<str>, T: FromStr>(
+        &self,
+        key: &K,
+        warnings: &mut RoadWarnings,
+    ) -> Option<T> {
+        self.get(key).and_then(|val| match val.parse::<T>() {
+            Ok(n) => Some(n),
+            Err(_) => {
+                warnings.push(RoadMsg::unsupported_tag(key.as_ref().to_owned(), val));
+                None
+            },
+        })
     }
 
     /// Return if tags key has value,
--- a/rust/osm2lanes/src/tag/mod.rs
+++ b/rust/osm2lanes/src/tag/mod.rs
@@ -93,22 +93,23 @@ impl Tags {
     }
 
     /// Get the value for the given key and parse it into T.
-    /// Add a RoadMsg::Unsupported if parsing fails.
+    /// Add a `RoadMsg::Unsupported` if parsing fails.
     // TODO: check if interface should be moved to transform
     pub fn get_parsed<K: AsRef<str>, T: FromStr>(
         &self,
         key: &K,
         warnings: &mut RoadWarnings,
     ) -> Option<T> {
-        self.get(key).and_then(|val| match val.parse::<T>() {
-            Ok(n) => Some(n),
-            Err(_) => {
+        self.get(key).and_then(|val| {
+            if let Ok(n) = val.parse::<T>() {
+                Some(n)
+            } else {
                 warnings.push(TagsToLanesMsg::unsupported_tag(
                     key.as_ref().to_owned(),
                     val,
                 ));
                 None
-            },
+            }
         })
     }
 
--- a/rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
+++ b/rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
@@ -1,0 +1,288 @@
+use crate::locale::{DrivingSide, Locale};
+use crate::tag::{TagKey, Tags};
+use crate::transform::tags_to_lanes::{Infer, Oneway, RoadBuilder};
+use crate::transform::{RoadWarnings, TagsToLanesMsg};
+
+const LANES: TagKey = TagKey::from("lanes");
+const ONEWAY: TagKey = TagKey::from("oneway");
+
+#[derive(Debug)]
+pub(in crate::transform::tags_to_lanes) struct Counts {
+    pub lanes: Infer<usize>,
+    pub forward: Infer<usize>,
+    pub backward: Infer<usize>,
+    pub both_ways: Infer<usize>,
+}
+
+impl Counts {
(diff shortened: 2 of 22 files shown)


droogmic:
I think for this to be readable I would have to split it, but that will take some more time :(

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
```diff
@@ -0,0 +1,375 @@
...
+use crate::road::Designated;
+use crate::tag::{TagKey, Tags};
+use crate::transform::tags_to_lanes::{Infer, Oneway, RoadBuilder};
+use crate::transform::{RoadWarnings, TagsToLanesMsg};
+
+const LANES: TagKey = TagKey::from("lanes");
+const ONEWAY: TagKey = TagKey::from("oneway");
+
+#[derive(Debug)]
+pub(in crate::transform::tags_to_lanes) struct Counts {
```
dabreegster, line 11: Ooh, never seen this fanciness before. Curious if `pub(super)` would be simpler? It could be a refactoring thing though; relative paths are more annoying when moving code around

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
```diff
@@ -0,0 +1,375 @@
...
+    pub(in crate::transform::tags_to_lanes) fn new(
+        tags: &Tags,
+        road: &RoadBuilder,
+        locale: &Locale,
+        warnings: &mut RoadWarnings,
+    ) -> Self {
+        // TODO, crosscheck access against tagged
+        let access = Self::get_counts_by_access(tags, warnings);
+
+        // Pre-existing lanes
```
dabreegster, line 33: I'm a little bit lost without extra context on _when_ this code runs, because it must be after some work has been done. Would it be worth a quick mention in the docstring?

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
```diff
@@ -0,0 +1,375 @@
...
+                ),
+                backward: tagged_backward.map_or(Infer::Default(pre_backward), Infer::Direct),
+                both_ways,
+            }
+        } else {
+            // Not oneway
+            match (tagged_lanes, tagged_forward, tagged_backward) {
+                (Some(l), Some(f), Some(b)) => {
+                    if l != f
+                        .checked_add(b)
```
dabreegster, line 136: Why `checked_add` and an unwrap? Are these just small `usize`s?

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
```diff
@@ -0,0 +1,375 @@
...
+                    ),
+                    forward: Infer::Direct(f),
+                    backward: Infer::Direct(b),
+                    both_ways,
+                },
+                (Some(l), Some(f), None) => Self {
+                    lanes: Infer::Direct(l),
+                    forward: Infer::Direct(f),
+                    backward: Infer::Calculated(
+                        l.checked_sub(f)
```
dabreegster, line 170: (`checked_sub` here does make sense, though I don't think we want to panic if something says `lanes=2, lanes:backward=5`)

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
```diff
@@ -0,0 +1,375 @@
...
+                            backward: Infer::Direct(remaining_lanes.checked_sub(half).unwrap()),
+                            both_ways,
+                        }
+                    }
+                },
+                (None, _, _) => {
+                    // Without the "lanes" tag, assume one normal lane in each dir
+                    // Adding any bus:lanes as needed.
+                    const DEFAULT: usize = 1;
+                    let tagged_bus_lanes: Option<usize> =
```
dabreegster, line 233: I'm a bit lost between this and #111 -- I thought we had a more consolidated struct just to handle bus stuff, and we would reconcile / merge the results of that here?

dabreegster (commented):
I see @BudgieInWA continuing to iterate on the other PR; I imagine as soon as we get the bulk of this approach checked in, there will be less confusing merge conflicts for the both of you...

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
droogmic: yeah, maintaining `pub(super)` becomes painful, especially when you need to refactor it to become `pub(super::super)`

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
droogmic: yeah, clippy doesn't like raw `+`s, but I think those are better than this, so I will revert

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
droogmic: I will add a TODO + expect for now

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
droogmic: I go a different route, for better or worse :/

droogmic:
> I imagine as soon as we get the bulk of this approach checked in, there will be less confusing merge conflicts for the both of you...

yeah, I am trying to keep getting small stuff in so that we reduce the size of the open WIP

BudgieInWA:
I have fixed the logic errors in #111, so all that is left is another merge of main, which I am doing right now. I feel that #111 has all that it needs to be merged, and would prefer that we do that instead of going ahead with this PR.

droogmic:
> I have fixed the logic errors in #111, so all that is left is another merge of main, which I am doing right now. I feel that #111 has all that it needs to be merged, and would prefer that we do that instead of going ahead with this PR.

maybe, but can you comment on how I handled the bus lanes.
I think this is actually a simpler, more natural, and more extensible way of handling these lanes? I will review your PR as well.

droogmic:
@BudgieInWA Note: I was actually made this whilst trying to merge your changes, but ended up wanting to simplify your work  to get it in in smaller section. Most code here is modified from yours.

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
```diff
@@ -0,0 +1,366 @@
...
+    fn get_counts_by_access(tags: &Tags, warnings: &mut RoadWarnings) -> Self {
+        const LANES: [TagKey; 2] = [TagKey::from("bus:lanes"), TagKey::from("psv:lanes")];
+        const FORWARD: [TagKey; 2] = [
+            TagKey::from("bus:lanes:forward"),
+            TagKey::from("psv:lanes:forward"),
+        ];
+        const BACKWARD: [TagKey; 2] = [
+            TagKey::from("bus:lanes:backward"),
+            TagKey::from("psv:lanes:backward"),
+        ];
```
BudgieInWA, line 280: Every instance of `<tag>:lanes` and `<tag>:lanes:{forward,backward}` should be checked, according to https://wiki.openstreetmap.org/wiki/Lanes.

If this calculation remains as part of `lanes` scheme parsing, then I would expect it to look at `*:lanes`, rather than a specific list of tags.

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
```diff
@@ -0,0 +1,366 @@
...
+                    l.designated.some(),
+                    Some(Designated::Bus | Designated::Motor)
+                )
+            })
+            .count();
+
+        // The tags for this schema (which we will validate).
+        let tagged_lanes: Option<usize> = tags
+            .get_parsed(&LANES, warnings)
+            .or_else(|| access.lanes.some());
```
BudgieInWA, line 64: This is no longer always the tagged value for the lanes key ("`tagged_lanes`"), but sometimes an implied count from access tags, but we can't tell.

rust/osm2lanes/src/tag/mod.rs
```diff
@@ -90,6 +92,27 @@ impl Tags {
         self.map.get(k.as_ref()).map(String::as_str)
     }
 
+    /// Get the value for the given key and parse it into T.
+    /// Add a `RoadMsg::Unsupported` if parsing fails.
+    // TODO: check if interface should be moved to transform
```
BudgieInWA, line 97: Yes, `get_parsed` exists for the convenience of things like `let tagged_lanes: Option<usize> = tags.get_parsed(&LANES, warnings)`, and should live in the `transform` module somewhere.

BudgieInWA (commented):
I think this evolution on the idea of parsing each scheme separately is moving in the wrong direction. More pre-conditions and "magic" is hidden away inside `Counts::new`, where I think we should be attempting to encapsulate tag parsing in a clear API and bringing the inter-dependencies out into the light, at some higher level.

If we have small scoped tag parsing methods answering questions like "what do the `lanes{,:forward,:backward}` tags say, (are they well formed and internally consistent)", then the top level `tags_to_lanes` can get on with adding things to the road builder and looking at the different schemes in just the right order.

When my `Scheme::new` functions where gaining lots of arguments, I too was tempted to replace them with a single `RoadBuilder` or something similar, but I thought it was important that the `Scheme` constructors be very clear about what their dependencies are, and that they are only other pre-parsed `Scheme`s. I now feel that these constructors should do *even less* than I had thought, but definitely not more. I will elaborate on this vision in #111.

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
droogmic: oh true, I didn't think of doing `*:lanes` globbed parsing. we don't have a facility to do that so that can be easily added later.

rust/osm2lanes/src/transform/tags_to_lanes/lane_count.rs
droogmic: indeed, so I live all the "hacks" in this monolithic function. this is the part I would totally refactor over time.

BudgieInWA: If that is the case, I think don't name it `Counts::new`, name it `calculate_and_guess_lane_counts(tags, road, ...) -> Counts` or something fittingly "hacky" :)

I found that converting to `Infer` as soon as possible, then using `.some()` if you don't care what you're getting, otherwise you can be smart about reconciling a default with a calculated value etc. I decided against chaining the logic like this, because it turns out if you are combining two sources of the same value, you want to do validation anyway, and having `Infer`s for the is useful.

PR #135, opened 2022-04-18
Rust: Rework Website

droogmic:
The way ID is now shown when selected on slippy map.

### DIFF
--- a/rust/osm2lanes-web/src/map.rs
+++ b/rust/osm2lanes-web/src/map.rs
@@ -9,10 +9,12 @@ use web_sys::{Element, HtmlElement, Node};
 use yew::prelude::*;
 use yew::Html;
 
+use crate::Msg as WebMsg;
+
 #[allow(clippy::large_enum_variant)]
 pub enum Msg {
     MapClick(LatLng),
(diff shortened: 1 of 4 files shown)


rust/osm2lanes-web/src/map.rs
```diff
@@ -9,10 +9,12 @@ use web_sys::{Element, HtmlElement, Node};
 use yew::prelude::*;
 use yew::Html;
 
+use crate::Msg as WebMsg;
+
 #[allow(clippy::large_enum_variant)]
 pub enum Msg {
     MapClick(LatLng),
-    MapUpdate(Tags, Locale, Vec<LatLon>),
+    MapUpdate(String, Tags, Locale, Vec<LatLon>),
```
dabreegster, line 17: I'd vote for switching to ```MapUpdate { way_id: String, tags: Tags, locale: Locale, points: Vec<LatLon> }``` when either the tuple list grows or one of the types isn't obvious

dabreegster (approved):
Nice UX change!

rust/osm2lanes-web/src/map.rs
droogmic: agreed, I did one but not the other :/

PR #141, opened 2022-04-23
Rust: Busway Scheme Refactor

droogmic:
In an attempt to make @BudgieInWA 's PR smaller.

### DIFF
--- a/rust/osm2lanes/src/transform/tags_to_lanes/modes/bus.rs
+++ b/rust/osm2lanes/src/transform/tags_to_lanes/modes/bus.rs
@@ -1,5 +1,5 @@
 use crate::locale::Locale;
-use crate::road::{Designated, Direction};
+use crate::road::Designated;
 use crate::tag::{TagKey, Tags};
 use crate::transform::tags_to_lanes::access_by_lane::Access;
 use crate::transform::tags_to_lanes::{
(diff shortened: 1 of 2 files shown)


rust/osm2lanes/src/transform/tags_to_lanes/modes/bus.rs
```diff
@@ -59,85 +59,179 @@ pub(in crate::transform::tags_to_lanes) fn bus(
...
-            "opposite_lane",
-        )
-        .into());
+
+    impl Scheme {
+        #[allow(clippy::unnecessary_wraps)]
+        pub(in crate::transform::tags_to_lanes) fn from_tags(
+            tags: &Tags,
+            locale: &Locale,
+            oneway: Oneway,
```
dabreegster, line 113: Takes a oneway and then calculates it below?

dabreegster (commented):
Seems like a step in the right direction and preserving the spirit of the other PR, but big warning, I'm rapidly losing the ability to keep up with what's happening here, so my feedback is not great. :\

rust/osm2lanes/src/transform/tags_to_lanes/modes/bus.rs
droogmic: the oneway coming in is for all vehicles, so I need to first check if there is a value specific for busses.

Now that you mention it, the bus-specific preprocessing should maybe happen one abstraction layer further out, but we can do that later.

droogmic:
> Seems like a step in the right direction and preserving the spirit of the other PR, but big warning, I'm rapidly losing the ability to keep up with what's happening here, so my feedback is not great. :\

Yeah, I am just looking for ways to help @BudgieInWA close out his work.

I really want to focus on fixing the widths of small residential roads, but that is mostly dependent on fixing the lane calculation part.

rust/osm2lanes/src/transform/tags_to_lanes/modes/bus.rs
dabreegster: Ah, I didn't read carefully enough. What if we name these `road_oneway` and `bus_oneway`?

BudgieInWA:
> Yeah, I am just looking for ways to help @BudgieInWA close out his work.
> 
> I really want to focus on fixing the widths of small residential roads, but that is mostly dependent on fixing the lane calculation part.

Thank you, you've done a great job of taking each idea and putting it into context, so that we tackle them independently.

PR #142, opened 2022-04-23
Rust: Update NL Autoweg Width

droogmic:
Referencing: https://puc.overheid.nl/rijkswaterstaat/doc/PUC_125514_31/ section 4.2.5

### DIFF
--- a/rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
+++ b/rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
@@ -11,6 +11,7 @@ use crate::transform::error::{RoadError, RoadWarnings};
 use crate::transform::RoadFromTags;
 
 mod error;
+use celes::Country;
 pub use error::TagsToLanesMsg;
 
 mod access_by_lane;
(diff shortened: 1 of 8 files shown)


droogmic:
image

This road is, as far as I can tell, the first fully correct representation of a road!

rust/osm2lanes/src/transform/tags_to_lanes/mod.rs
```diff
@@ -277,12 +278,27 @@ impl RoadBuilder {
...
+                if c == &Country::the_netherlands() {
+                    for lane in road.lanes_ltr_mut(locale) {
+                        lane.width = Width {
+                            target: Infer::Default(Metre::new(3.35)),
+                            ..Default::default()
+                        }
+                    }
+                }
+            }
+        }
```
droogmic, line 299: an obvious hack... so during some refactors I will look for nicer ways to house this code.

dabreegster: Better to start with something! I like how code is organized by part of the road now. Maybe locale-specific widths should act like a cross-cutting concern and be housed in their own module, and act as a post-processing step. Then we can check if we're in the Netherlands, and if so, walk through everything and add detail. But probably best to have at least one other locale-specific example before trying to refactor

dabreegster (approved):
Awesome to see end-to-end!

PR #143, opened 2022-04-23
Rust: Access to Spec

droogmic:
A proposal on how to add access definitions to the spec.

### DIFF
--- a/rust/osm2lanes/src/road/lane.rs
+++ b/rust/osm2lanes/src/road/lane.rs
@@ -3,6 +3,7 @@ use serde::{Deserialize, Serialize};
 use super::Markings;
 use crate::locale::Locale;
 use crate::metric::{Metre, Speed};
+use crate::tag::Access as AccessValue;
 
 /// A single lane
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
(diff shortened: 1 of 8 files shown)


rust/osm2lanes/src/road/lane.rs
```diff
@@ -158,3 +161,20 @@ impl Printable for Direction {
         }
     }
 }
+
+/// Access by vehicle type
+/// Types as defined in <https://wiki.openstreetmap.org/wiki/Key:access#Land-based_transportation>
```
dabreegster, line 166: Are we not helping dogs figure out where they're allowed to be? ;)

rust/osm2lanes/src/road/lane.rs
```diff
@@ -158,3 +161,20 @@ impl Printable for Direction {
...
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub struct Access {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    foot: Option<AccessValue>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    bicycle: Option<AccessValue>,
```
dabreegster, line 173: The only case that may come up from A/B Street in the medium-term is bicycle=dismount. But just something to keep in mind, this all seems fine to start

dabreegster (approved):

PR #147, opened 2022-04-25
Remove dying languages

droogmic:
Remove Kotlin and Python, they can be put back if need be.


README.md
```diff
@@ -1,159 +1,6 @@
-# osm2lanes
+# Development
 
-This project transforms OpenStreetMap tags to a specification of lanes on a
```
dabreegster, line 3: Too much got deleted here and not moved?

dabreegster (approved):
Minus part of the README vanishing, looks good to me. Thanks for the cleanup!

I was thinking of suggesting this when the Kotlin tests were failing the other day. Part of me wanted to set up JVM and Python bindings first, but no need to do this now.

README.md
droogmic: oh, my move overwrote this, didn't notice :(

droogmic:
The benchmark fails because the paths have changed, it will fix itself when the target also has the same structure

dabreegster:
> The benchmark fails because the paths have changed, it will fix itself when the target also has the same structure

I'm never opposed to a force push ;)

PR #158, opened 2022-04-28
Refactor Non-motorized

droogmic:
Changes to Infer, changes to lane counts.
Currently makes a dangerous assumption on roadbuilder state, will need to be reworked later


droogmic:
This PR is a bit big, feel free to ask me to split it.

osm2lanes/src/transform/tags_to_lanes/counts.rs
```diff
@@ -208,6 +186,59 @@ impl Counts {
...
+                        }
+                    } else {
+                        Self {
+                            lanes: Infer::Default(1),
+                            forward: Infer::Default(0),
+                            backward: Infer::Default(0),
+                            both_ways: Infer::Default(1),
+                        }
+                    }
+                },
```
droogmic, line 241: not sure this change was necessary

osm2lanes/src/locale.rs
```diff
@@ -73,6 +73,28 @@ impl Locale {
         }
     }
 
+    /// Highway type with no `lanes=*` has one bidirectional lanes or two directions lanes
```
dabreegster, line 76: Having trouble parsing this. "has one lane in each direction"? "one bidirectional lane" to me means "1 lane, that can go both directions"

osm2lanes/src/transform/tags_to_lanes/infer.rs
```diff
@@ -0,0 +1,117 @@
+use super::TagsToLanesMsg;
```
dabreegster, line 1: Going forward, if there's a PR moving stuff around, can we _just_ do that step first? Much easier to review mechanical refactors separate from ones that change something interesting. Because git doesn't pick up the rename well, I don't really know if something in here changed

osm2lanes/src/transform/tags_to_lanes/modes/non_motorized.rs
```diff
@@ -1,133 +1,38 @@
...
-            road: Road {
-                lanes: vec![Lane::foot(locale)],
-                highway: road.highway.clone(),
-            },
-            warnings: RoadWarnings::default(),
-        }));
-    }
-    // Otherwise, there'll always be a bike lane.
-
-    let mut forward_side = vec![Lane::forward(Designated::Bicycle, locale)];
```
dabreegster, line 94: I think this was essentially dead code -- if the road allowed a mix of walking and cycling, another part already would've filled out cycle lanes

dabreegster (approved):
Thanks for detangling this old mess

osm2lanes/src/locale.rs
droogmic, line 76:
```suggestion
    /// Highway type with no `lanes=*` has:
    /// - (false) one lane with travel in both directions or 
    /// - (true) two lanes with travel in opposite directions
```

osm2lanes/src/transform/tags_to_lanes/infer.rs
droogmic: nothing meaningful, but agreed, I will make more of the refactoring PRs I made before

osm2lanes/src/locale.rs
```diff
@@ -73,6 +73,30 @@ impl Locale {
         }
     }
 
+    /// Highway type with no `lanes=*` has:
+    /// - (false) one lane with travel in both directions or 
+    /// - (true) two lanes with travel in opposite directions
```
droogmic, line 78:
```suggestion
    /// Highway type with no `lanes=*` has:
    /// - (false) one lane with travel in both directions or
    /// - (true) two lanes with travel in opposite directions
```

Issue #152, opened 2022-04-27
attempt to subtract with overflow


dabreegster:
cargo run way 240294912
https://www.openstreetmap.org/way/240294912

osm2lanes/src/transform/tags_to_lanes/lane.rs:205:33

droogmic:
`lanes:both_ways=2`, an odd one for sure, I don't even know what that means.

dabreegster:
I think it's just a typo. https://wiki.openstreetmap.org/wiki/Key:lanes:both%20ways?uselang=en doesn't mention any other values. By and large, 1 is the value used https://taginfo.openstreetmap.org/keys/lanes%3Aboth_ways#values.

Should we
1) Throw an error
2) Assume any value besides "1" is a typo for "1" and emit a warning?
3) Assume any value besides "1" is a typo and not add anything?

I'll start with option 1 and at least not panic.

droogmic:
My PR now implements the 3rd option, warning and ignore any value that
isn't 1.

dabreegster:
Awesome, thank you! I'll add a test for this

PR #169, opened 2022-04-30
Add a test for #152

dabreegster:
If this test is enabled, we panic

### DIFF
--- a/data/tests.yml
+++ b/data/tests.yml
@@ -271,6 +271,39 @@
           - style: solid_line
             color: white
 
+### Mis-tagged roads
+
+- way_id: 240294912
+  description: Invalid lanes:both_ways value
+  rust:
+    expect_warnings: true
+  tags:
+    highway: "primary"
+    lanes: "4"
+    lanes:both_ways: "2"
+    sidewalk: "both"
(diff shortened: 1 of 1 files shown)


data/tests.yml
```diff
@@ -271,6 +271,36 @@
...
+  tags:
+    highway: "primary"
+    lanes: "4"
+    lanes:both_ways: "2"
+    sidewalk: "both"
+  driving_side: left
+  ISO 3166-2: GB-NET
+  road:
+    highway: primary
+    lanes:
```
dabreegster, line 289: I am actually not sure what the correct interpretation of these tags is...

droogmic: maybe we need some kind of "don't care" output? just to check that we don't panic

likewise I intend to and an error checking possibility in the output

droogmic (approved):

data/tests.yml
dabreegster: That'd be nice, will keep it in mind. I enabled the test and made the expectations match the output -- which is kind of half-reasonable in this case. Just ignore the bogus both_ways value.

Issue #136, opened 2022-04-18
Mitigation for more than one bus lanes scheme used


dabreegster:
By far the most common problem I'm seeing with real data is more than one bus lanes scheme used. One example is https://www.openstreetmap.org/way/679636490. Glancing at this, I can't help feeling like though it uses two schemes, they both agree -- there's exactly 1 backwards lane and it's bus-only.

For #71, I'll probably be tactical in the short-term, detect these cases, and just arbitrarilyish rip out keys for one of the schemes. But I had another idea -- if we could parse each bus scheme independently, clone the road builder, and "try out" both schemes, then we could check if the results match. If so, just a warning about redundant tags could suffice -- no need to totally give up?

droogmic:
I actually already allow busway + either bus:lanes or lanes:bus in https://github.com/a-b-street/osm2lanes/pull/134/files#diff-22ad929b7a488fd33b93d20054e80bb72678cebda592203907bb58188c8608b7

allowing for both bus:lanes and lanes:bus may be possible, but we have yet to implement lanes:bus, so it doesn't really matter yet

BudgieInWA:
Each scheme describes a slightly different aspect of the road, and they should be reconciled. I think the `XScheme` types in #111 allow for this reconciliation to be implemented in a sensible way. If the tags don't agree with each other, then they are invalid and we have to decide which tags to ignore, which we could do simply by preferring one schema over another, or we could get as fancy as we like. I feel like a valuable usecase for osm2lanes is detecting these incompatibilities and suggesting the simplest fix, so maybe we do end up getting pretty fancy about it in the long run.

The situation is exactly the same with `lanes` and `lanes:{forward,backward}`: If `forward` and `backward` are provided, `lanes` is redundant, but it is convention to include it for less sophisticated consumers. If the tags disagree, we would have to choose one tag to ignore in order to produce an answer.

- `lanes:bus` is the `lanes` tag narrowed in scope to just buses, so "how many bus lanes are there". (On a twoway road, I would expect `lanes:bus:{forward,backward}` instead).
- `bus:lanes` is the `bus` access tag, but on a per lane basis, so "what is the access of each lane, left to right". (Again, in my experience on a twoway road, I would expect `bus:lanes:{forward,backward}` instead).
- `busway` is an older scheme that is ambiguous to the number of lanes in each direction and the position of those lanes. I would discourage it's use in `lanes_to_tags` unless the user specifically requests it.

[...]

PR #173, opened 2022-04-30
Add Test For 679636490

droogmic:
Combining bus:lanes and lanes:bus

### DIFF
--- a/data/tests.yml
+++ b/data/tests.yml
@@ -309,8 +309,7 @@
 # incorrect speed limit
 - way_id: 102917976
   description: Road with impossible speed limit
-  rust:
-    false
+  rust: false
     #expect_warnings: true
     #separator: true
   tags:
--- a/data/tests.yml
+++ b/data/tests.yml
@@ -309,7 +309,8 @@
 # incorrect speed limit
 - way_id: 102917976
   description: Road with impossible speed limit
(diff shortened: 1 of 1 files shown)


droogmic:
Taken from discussion in https://github.com/a-b-street/osm2lanes/issues/136

data/tests.yml
```diff
@@ -309,7 +309,8 @@
 # incorrect speed limit
 - way_id: 102917976
   description: Road with impossible speed limit
-  rust: false
```
dabreegster, line 312: Ah, do you have a process for auto-formatting this file?

dabreegster (approved):

data/tests.yml
droogmic: I... don't know.
VS code sometimes fiddles with the yaml I write but I don't know when or why :/

It's definitely not something I would know how to integrate with the CI (without investigating)

data/tests.yml
```diff
@@ -309,7 +309,8 @@
...
+  rust:
+    false
```
droogmic, line 313:
```suggestion
  rust: false
```

data/tests.yml
droogmic: I reverted that line for now

PR #185, opened 2022-05-03
Add Examples To Website

droogmic:
Add a convenient dropdown to website for good examples

### DIFF
--- a/osm2lanes-web/src/control.rs
+++ b/osm2lanes-web/src/control.rs
@@ -1,15 +1,30 @@
 use std::cell::RefCell;
+use std::collections::BTreeMap;
 use std::rc::Rc;
 
-use osm2lanes::locale::{Country, DrivingSide};
+use gloo_timers::callback::Timeout;
+use osm2lanes::locale::{Country, DrivingSide, Locale};
+use osm2lanes::test::{get_tests, TestCase};
 use web_sys::{Event, FocusEvent, HtmlInputElement, HtmlSelectElement, KeyboardEvent, MouseEvent};
 use yew::{html, Callback, Component, Context, Html, NodeRef, Properties, TargetCast};
 
-use crate::{Msg, State};
+use crate::{Msg as AppMsg, State};
+
+pub enum Msg {
+    Up(Box<AppMsg>),
+    FirstLazy,
+    Example(String),
+}
+
+impl From<AppMsg> for Msg {
+    fn from(msg: AppMsg) -> Self {
+        Self::Up(Box::new(msg))
+    }
+}
 
 #[derive(Properties, PartialEq)]
 pub struct Props {
-    pub callback_msg: Callback<Msg>,
+    pub callback_msg: Callback<AppMsg>,
     pub state: Rc<RefCell<State>>,
 }
 
@@ -17,6 +32,8 @@ pub struct Props {
 pub struct Control {
     textarea_input_ref: NodeRef,
     textarea_output_ref: NodeRef,
+    example: Option<String>,
+    examples: Option<BTreeMap<String, TestCase>>,
 }
 
 impl Component for Control {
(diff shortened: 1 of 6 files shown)


droogmic:
My vscode did something to the .css file, I don't know why :/

osm2lanes-web/src/control.rs
```diff
@@ -175,4 +254,14 @@ impl Component for Control {
...
     }
+
+    fn rendered(&mut self, ctx: &Context<Self>, first_render: bool) {
+        if first_render {
+            let handle = {
+                let link = ctx.link().clone();
+                Timeout::new(1, move || link.send_message(Msg::FirstLazy))
+            };
+            handle.forget();
+        }
```
droogmic, line 265: hacky...

osm2lanes/src/test.rs
```diff
@@ -122,14 +129,11 @@ impl std::fmt::Display for TestCase {
...
-    let tests: Vec<TestCase> =
-        serde_yaml::from_reader(BufReader::new(File::open("../data/tests.yml").unwrap()))
-            .expect("invalid yaml in data/tests.yml");
-    let tests: Vec<TestCase> = tests
-        .into_iter()
-        .filter(|test| test.test_enabled())
-        .collect();
+    let tests: Vec<TestCase> = serde_yaml::from_str(include_str!("../../data/tests.yml"))
+        .expect("invalid yaml in data/tests.yml");
+    let tests: Vec<TestCase> = tests.into_iter().filter(TestCase::test_enabled).collect();
```
droogmic, line 136: we now include the tests at compile time (File::open not supported in wasm)

downside: every test change triggers a recompile.
necessary evil?

I could make a crate dedicated to this, or have platform specific implementation, but both seem worse.

osm2lanes-web/src/control.rs
```diff
@@ -1,36 +1,84 @@
...
 use web_sys::{Event, FocusEvent, HtmlInputElement, HtmlSelectElement, KeyboardEvent, MouseEvent};
 use yew::{html, Callback, Component, Context, Html, NodeRef, Properties, TargetCast};
 
-use crate::{Msg, State};
+use crate::{Msg as AppMsg, State};
+
+pub enum Msg {
+    Up(Box<AppMsg>),
+    FirstLazy,
+    Example(String),
```
dabreegster, line 16: Just checking my understanding here... `Up` refers to the parent component, `FirstLazy` is a trick to lazily load up test examples, and `Example` is when a particular example has been set

osm2lanes-web/src/control.rs
dabreegster: Why don't we do all the work of loading examples up-front and avoid the indirection? Is it slow enough to block the page load or something?

osm2lanes/src/test.rs
```diff
@@ -34,6 +32,9 @@ pub struct TestCase {
     pub comment: Option<String>,
     pub description: Option<String>,
 
+    // Examplary
```
dabreegster, line 35:
```suggestion
    // List as a named example in the web app
```
unless we don't want to tie it so explicitly to that one use. (But it is otherwise a bit unclear why `description` is different)

osm2lanes/src/test.rs
dabreegster: As long as this is feature-gated, seems fine to me. Test changes triggering a recompile really is necessary for the web app.

FWIW, there are even crates like https://github.com/Michael-F-Bryan/include_dir/ to shove more stuff into WASM at compile-time. I... have definitely included about 500MB of binary files before. :D

dabreegster (commented):
Seems fine to me, minus the necessity of lazily loading examples.

I'm less able to try this locally; main dev machine currently broken, and `cargo install trunk` very bizarrely broke...

osm2lanes-web/src/control.rs
droogmic: I noticed the chance, I think 50% of the page load is on deserializing the .yml

osm2lanes-web/src/control.rs
droogmic: yeah, naming is hard

dabreegster (approved):

droogmic:
After:
image

droogmic:
Before: 
image

droogmic:
You can really see how the yaml deserialization consumes about 1.1s of load time...

droogmic:
https://gtmetrix.com/reports/a-b-street.github.io/gVqO8od3/

dabreegster:
Grade A, nice! Deferring the YAML load makes perfect sense to me then.

By the way, if you happen to be familiar with how to handle this kind of problem in web browsers, do you know of the "proper" way to load/parse large blobs of data without freezing things up? Downloading a large file and showing a progress bar is possible to do async by fetching in chunks. But parsing it blocks the CPU. Do any/many web apps have a nicer UX? The only two ideas to workaround are to try web workers (which don't seem to be easy to manage from Rust, though I've spotted some more recent crates that maybe help) or to use some kind of streaming serialization that can limit itself to a real deadline, yield, and let the caller update a progress bar or give the user control to interact with the page. For the latter idea, serde doesn't seem to help directly. But if the large blob is a list of stuff, then one approach is to just try and deserialize one item at a time. I haven't seen any real examples doing this though...

(All of that's an aside; this PR looks fine to me -- great new feature)

droogmic:
> By the way, if you happen to be familiar with how to handle this kind of problem in web browsers, do you know of the "proper" way to load/parse large blobs of data without freezing things up? Downloading a large file and showing a progress bar is possible to do async by fetching in chunks. But parsing it blocks the CPU. Do any/many web apps have a nicer UX? The only two ideas to workaround are to try web workers (which don't seem to be easy to manage from Rust, though I've spotted some more recent crates that maybe help) or to use some kind of streaming serialization that can limit itself to a real deadline, yield, and let the caller update a progress bar or give the user control to interact with the page. For the latter idea, serde doesn't seem to help directly. But if the large blob is a list of stuff, then one approach is to just try and deserialize one item at a time. I haven't seen any real examples doing this though...

I'm not an expert, basically a no-stack developer... I do think webworkers are the way, I was just waiting for good integration with trunk... https://github.com/thedodd/trunk/pull/285

Oh it's done. OK, maybe I will try to make a web-worker as an example then...

droogmic:
it's basically https://github.com/yewstack/yew/tree/master/examples/web_worker_fib

droogmic:
looks simple enough :)

droogmic:
Still OK after: https://gtmetrix.com/reports/a-b-street.github.io/mX4s9uCK/

Issue #161, opened 2022-04-28
Cycleway lanes on oneway streets


hungerburg:
A recent ping reminded me to update my local A/B-Streets install. Browsing around my city in A/B-Street, which is always a pleasure, I got to see some bidirectional cycleways, which in fact are single direction only. Eg. https://www.openstreetmap.org/way/25745877 has a lane for cyclists contra-flow motorists. Its tagging looks like polished with StreetComplete, still I consider the tagging accurate and complete, if not over-complete.

In A/B-Street, the "opposite-lane" (an old-fashioned osm jargon) for cyclists renders as a bidirectional cycle-way, where in reality, it is just a narrow cycle-only-lane contra the one-way. Cyclists going the motorist-one-way direction are to use the carriageway just as usual.

The OSM-Wiki is silent about directionality in such cases. I asked on the new forum, but only got on response from somebody, sharing the same background that I operate on ;) Is this different in e.g. Seattle? Or, are there just no tiles/presets for that? Or …

hungerburg:
One cannot have a conversation on openstreetmap, so it seems, without competing ideas how to map the same stuff in different ways. Luckily, consumers need not take part in endless discussions. Unfortunately, consumers need to support different schemes.

The scheme applied on location indeed is very map-painter friendly. Just draw a blue casing on the side of the osm-way, that the cycle lane is said to be on and add a blue contra arrow to the one-way arrow of the base highway. It is also convenient to provide meaningful turn-by-turn instructions from a router.

Things get more complicated, when working on lane level, with lanes that always have a single direction. From what I can tell, cycleways in A/B-Street are of this kind.

Looking at the osm2lanes calculation, it seems to me, that it infers "cycleway:left:oneway=no" from "oneway:bicycle=no". I think, this step goes too far. I see a bit of head scratching ahead, to get the correct direction of cycle lanes tagged under this scheme. No idea, if this is even worth the effort. Though it is wrong, and so looks strange, for simulation purposes, it does not really matter much. I might spend a bit more thinking on how to solve this.

hungerburg:
I see, that the text in the left-box on osm2lanes is editable. Shortened version of fetched tags gives me a bidirectional cycle lane:

```
cycleway:left=lane
cycleway:right=no
highway=residential
oneway=yes
oneway:bicycle=no
sidewalk=both
```

When I add the line  below, I get an arrow on the cycle lane in direction of main traffic:

`cycleway:left:oneway=yes`

[...]

PR #189, opened 2022-05-07
Add Tests For cycleway:BACKWARD=lane

droogmic:


droogmic:
For https://github.com/a-b-street/osm2lanes/issues/161

data/tests.yml
```diff
@@ -730,6 +730,85 @@
...
+    lanes:
+      - type: travel
+        designated: foot
+      # - type: separator
+      #   markings:
+      #     - style: kerb
+      - type: travel
+        direction: backward
+        designated: bicycle
+      - type: separator
```
dabreegster, line 759: Wait, why're some of the separators enabled and some commented out?

dabreegster (approved):
I got lost following the discussion in the issue a while ago... but tests locking in our declaration of correctness are good

data/tests.yml
droogmic: that one is clear, the schema for kerbs is still up in the air...

reminds me that I need to fix that :P

PR #192, opened 2022-05-09
Add Semantic Separator To Spec

droogmic:
Added as flat field, completeness and naming is still to be decided.

### DIFF
--- a/osm2lanes/src/road/separator.rs
+++ b/osm2lanes/src/road/separator.rs
@@ -4,6 +4,20 @@ use super::Printable;
 use crate::locale::Locale;
 use crate::metric::Metre;
 
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, PartialEq, Serialize, Deserialize)]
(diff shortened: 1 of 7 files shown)


osm2lanes/src/road/separator.rs
```diff
@@ -4,6 +4,20 @@ use super::Printable;
 use crate::locale::Locale;
 use crate::metric::Metre;
 
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum Semantic {
```
dabreegster, line 10: Maybe after we settle on including this in the output, it would make sense to use the same enum internally and in the public API. But no need to rush that decision.

I also remembered https://wiki.openstreetmap.org/wiki/Proposed_features/cycleway:separation#Typical_values for future ideas here

dabreegster (approved):

PR #202, opened 2022-05-17
Remove TagsWrite

droogmic:
A common newbie rust mistake: overuse traits


dabreegster (approved):
A common reviewer mistake: fail to question things sufficiently. :)

Issue #78, opened 2022-02-10
Tags Data Structure


droogmic:
https://wiki.openstreetmap.org/wiki/Tags

# Requirements

## Key

A key type, at its simplest it could be a `&'static str` or `String`. See also: <https://lib.rs/crates/kstring>.

- No heap usage when using `&'static str`s.
- `Add` trait implemented to support `:` joined keys (i.e. `A + B + C == "a:b:c"`)

## Value

A value type, at its simplest it could be a `&'static str` or `String`. 
Trait implemented to support `;` joined keys.
The result should also resemble an iterable.
Example: `[A, B, C].into() == "a;b;c"`.

## Mapping

Support interaction mimicking a `HashMap` or `BTreeMap` `<K, V>`, for example:

- `fn get<Q>(&self, key: &Q) -> Option<V>`
- `fn is<Q, U>(&self, key: &Q, value: &U) -> bool`
- `fn is_any<Q, W>(&self, key: &Q, values: &W) -> bool`

Arguments of type `Q` should be possible directly from a `&'static str`, .e.g. `tags.get("highway")`.

## Tree

Support access as a tree structure, for `:` separated keys.
Example:
```
// foo:bar:baz=toto
tags.tree().get("foo").unwrap().get("bar").unwrap().is("baz", "toto")
```

## Performance Constraints

In addition to the constraints on the key (see kstring):

- The data must be stored once (not both as a tree and as a map, as is currently done)
- The data may not be stored as a naive map, as retrieving a key of `foo:bar` using `foo` will be O(N).
- Retrieving a multi-part key must not incur a runtime `:` split cost when not needed, i.e. `get("foo" + "bar")` must not allocate a `String("foo:bar")` which is then `.split_once(':')`.

# Deployment

For now, this should replace <https://github.com/a-b-street/osm2lanes/tree/main/rust/osm2lanes/src/tag>.
Once proven, a separate crate should be made.

droogmic:
for key types: https://github.com/epage/string-benchmarks-rs

> Suggestions:
> - Generally, String
> - If you deal mostly with string literals but want some flexibility (like [clap](https://github.com/clap-rs/clap/)), generally you'll want Cow<'static, str>
> - If a profiler says your strings are a problem:
>   Try different crates and settings for that crate out with a profiler
>   O(1) clones are important when doing a lot of clones. For one-off allocations, they are slower.
>   For short-lived programs, look into string interning

PR #203, opened 2022-05-20
OSM Tags only as Tree

droogmic:
An initial step towards a better tree structure.
Generics removed for simplicity
Successful tree-only data store

### DIFF
--- a/osm-tags/src/lib.rs
+++ b/osm-tags/src/lib.rs
@@ -45,14 +45,15 @@
     clippy::use_debug
 )]
 
+use std::borrow::Borrow;
 use std::collections::BTreeMap;
 use std::str::FromStr;
 
 mod key;
-pub use key::TagKey;
+pub use key::{TagKey, TagKeyPart};
 
-mod osm;
-pub use osm::{Highway, HighwayImportance, HighwayType, Lifecycle, HIGHWAY, LIFECYCLE, ONEWAY};
+// mod osm;
+// pub use osm::{Highway, HighwayImportance, HighwayType, Lifecycle, HIGHWAY, LIFECYCLE, ONEWAY};
 
 mod access;
 pub use access::Access;
@@ -69,7 +70,7 @@ impl From<String> for DuplicateKeyError {
 
 impl std::fmt::Display for DuplicateKeyError {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "duplicate tag key {}", self.0.as_str())
+        write!(f, "duplicate tag key {}", self.0.to_string())
     }
 }
 
@@ -82,12 +83,8 @@ impl std::error::Error for DuplicateKeyError {}
 // We often need to compare output directly, so cannot tolerate reordering
 //
 // TODO: fix this in the serialization by having the keys sorted.
-//
-// TODO: use only one of map or tree, with a zero-cost API for the other at runtime
-//
 #[derive(Clone, Debug, Default)]
 pub struct Tags {
-    map: BTreeMap<String, String>,
     tree: TagTree,
 }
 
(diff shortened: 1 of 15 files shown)


osm-tags/src/lib.rs
```diff
@@ -82,12 +83,8 @@ impl std::error::Error for DuplicateKeyError {}
 // We often need to compare output directly, so cannot tolerate reordering
 //
 // TODO: fix this in the serialization by having the keys sorted.
-//
-// TODO: use only one of map or tree, with a zero-cost API for the other at runtime
-//
 #[derive(Clone, Debug, Default)]
 pub struct Tags {
-    map: BTreeMap<String, String>,
```
droogmic, line 90: this is the goal :)

droogmic:
I have gotten _slighlty_ closer to the goals outlined in https://github.com/a-b-street/osm2lanes/issues/78, but there is still a lot to be direct. I appear to have stumbled into quite a hard problem in the `get` and similar interfaces for the query type Q.

- if Q is a `&str`, static or otherwise, we should coerce this into an `iter::once`
- if Q is a `TagKey`, we should move it in, get the parts, and drop it
- if Q is a `&TagKey`, we should reference the parts.

The current PR achieves the first two, but not the third, as it triggers a clone into `TagKey` before referencing again.

I have tried the following:
- Into/From
- Borrowed
- AsRef
- IntoIterator

I did not try Cow, even though it might work, as it makes the decision at runtime when we shouldn't need that?

Summary: I need someone who knows more than me about rust to give advice, but I think I'll need to make a toy example of the problem first.

droogmic:
```
 ┌─────────┬──────────────────────────────────────────────────────────┬──────────────────┬─────────────────┬────────────┐
│ (index) │                           name                           │   baseDuration   │ changesDuration │ difference │
├─────────┼──────────────────────────────────────────────────────────┼──────────────────┼─────────────────┼────────────┤
│    0    │                    'tests/224637155'                     │ '**5.4±0.26µs**' │  '11.6±0.87µs'  │ '+1.1e+2'  │
│    1    │          'tests/380103730 Japanese Expressway'           │ '**7.1±0.35µs**' │  '13.9±0.79µs'  │   '+97'    │
│    2    │                    'tests/389654080'                     │ '**7.9±0.43µs**' │  '14.7±0.79µs'  │   '+85'    │
│    3    │         'tests/49207928 cycleway:BACKWARD=lane'          │ '**7.2±0.32µs**' │  '14.9±0.77µs'  │ '+1.1e+2'  │
│    4    │ 'tests/8591383 a bidirectional cycleway, oneway:bicycle' │ '**8.1±0.58µs**' │  '15.6±0.81µs'  │   '+93'    │
│    5    │              'tests/bus:lanes=designated|'               │ '**7.1±0.27µs**' │  '15.8±0.98µs'  │ '+1.2e+2'  │
│    6    │                   'tests/busway=lane'                    │ '**6.7±0.35µs**' │  '13.6±1.02µs'  │ '+1.0e+2'  │
│    7    │                  'tests/cycleway=lane'                   │ '**5.8±0.35µs**' │  '12.4±0.54µs'  │ '+1.1e+2'  │
│    8    │                'tests/sidewalk:right=yes'                │ '**5.3±0.25µs**' │  '11.8±0.63µs'  │ '+1.2e+2'  │
│    9    │                  'tests/sidewalk=both'                   │ '**5.6±0.32µs**' │  '12.0±0.72µs'  │ '+1.1e+2'  │
│   10    │                            ''                            │    undefined     │    undefined    │   '+NaN'   │
└─────────┴──────────────────────────────────────────────────────────┴──────────────────┴─────────────────┴────────────┘
```

performance has regressed across the board.

I slowed down the common path.
This might be enough to justify killing the PR

droogmic:
maybe I go the opposite direction, I remove the tree, and instead recompute it at runtime if/when needed.

There are currently very few uses of the "tree" API, basically only in `bus` when looking for `lanes:bus:*`, where we even know what the `*` might be...

dabreegster:
> There are currently very few uses of the "tree" API

It feels a slight bit over-engineered for the scale we've got -- 20 tags in a way would be rare. Even if we did linear search and `str.starts_with`, I bet performance would be OK. Way back in my comp sci undergrad, I remember using [tries](https://en.wikipedia.org/wiki/Trie) and I spot crates like https://crates.io/crates/qp-trie. But for so few entries, worth the complexity?

droogmic:
indeed, let me try again.

PR #204, opened 2022-05-21
Miscellaneous Map Fixes and Improvements

droogmic:


osm2lanes-web/src/map.rs
```diff
@@ -37,26 +38,32 @@ impl MapComponent {
...
     }
 }
 
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct MapOptions {
+    scroll_wheel_zoom: bool,
+}
+const MAP_OPTIONS: MapOptions = MapOptions {
+    scroll_wheel_zoom: false,
```
dabreegster, line 54: Aw, I kind of miss this. Double clicking to zoom feels a bit too similar to clicking to focus somewhere. But I get the tension between scrolling the page and map

dabreegster (approved):
Changes mostly feel smooth to me! One thing I noticed as I clicked around quickly trying to see the error messages show up is that multiple in-flight requests appear to sometimes override each other in slightly unexpected ways. I guess it's just the overpass lag; I'll click twice, see one change, then the second request finishes a few seconds later and updates everything a bit unexpectedly. Not sure how other apps solve this UX problem to clearly show cancelled/pending operations, will think about it a bit

osm2lanes-web/src/map.rs
droogmic: yeah, I wanted to try and get ctrl+scroll working, but I didn't see an option for that in leaflet

droogmic:
> multiple in-flight requests

I was about to make another PR to prevent that as well :)

osm2lanes-web/src/map.rs
droogmic: https://github.com/elmarquis/Leaflet.GestureHandling

that's what I want!
I couldn't find it because it is not part of leaflet natively...

PR #205, opened 2022-05-23
Refactor Tags Only Map

droogmic:
An attempt 2 at refactoring tags.

### DIFF
--- a/osm-tags/src/lib.rs
+++ b/osm-tags/src/lib.rs
@@ -127,7 +127,8 @@ impl Tags {
         V: Into<TagVal>,
     {
         let mut map = BTreeMap::default();
-        map.insert(key.into(), val.into()).unwrap();
+        let duplicate_val = map.insert(key.into(), val.into());
+        debug_assert!(duplicate_val.is_none());
         Self { map }
     }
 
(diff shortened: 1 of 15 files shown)


osm-tags/src/lib.rs
```diff
@@ -82,150 +88,154 @@ impl std::error::Error for DuplicateKeyError {}
...
 //
 // TODO: fix this in the serialization by having the keys sorted.
-//
-// TODO: use only one of map or tree, with a zero-cost API for the other at runtime
-//
 #[derive(Clone, Debug, Default)]
 pub struct Tags {
-    map: BTreeMap<String, String>,
-    tree: TagTree,
+    map: BTreeMap<TagKey, TagVal>,
```
droogmic, line 93: this was the goal...

osm2lanes/src/transform/tags_to_lanes/counts.rs
```diff
@@ -232,9 +232,9 @@ impl LanesDirectionScheme {
             })
             .map(|_v| {});
         Self {
-            total: tags.get_parsed(LANES, warnings),
-            forward: tags.get_parsed(LANES + "forward", warnings),
-            backward: tags.get_parsed(LANES + "backward", warnings),
+            total: tags.get_parsed(&LANES, warnings),
```
dabreegster, line 235: This seems like a slight regression in API niceness, but not by much. Working off limited brain capacity, it looks like this used to do some more cloning maybe

osm2lanes/src/transform/tags_to_lanes/modes/bus/mod.rs
```diff
@@ -61,15 +61,11 @@ pub(in crate::transform::tags_to_lanes) fn bus(
...
-        tags.tree().get("busway").is_some(),
-        tags.tree()
-            .get("lanes:bus")
-            .or_else(|| tags.tree().get("lanes:psv"))
-            .is_some(),
-        tags.tree()
-            .get("bus:lanes")
-            .or_else(|| tags.tree().get("psv:lanes"))
-            .is_some(),
+        !tags.pairs_with_stem("busway").is_empty(),
```
dabreegster, line 64: Some of the substantial change here. This API looks fine!

dabreegster (approved):
Difficult to review, but if it compiles and tests pass, then great. The benchmark changes look like noise to me.

osm2lanes/src/transform/tags_to_lanes/counts.rs
droogmic: yeah, maybe we can find a way to add it back later, but I really wrestled with getting this working nicely.
the interface also makes more sense this way (a reference, no internal cloning).

PR #212, opened 2022-05-25
Clippy on osm2lanes-web

droogmic:


osm2lanes-web/src/control.rs
```diff
@@ -52,11 +52,11 @@ impl Component for Control {
                 let examples: BTreeMap<_, _> = tests
                     .into_iter()
                     .filter_map(|t| {
-                        let example_name = t.example().map(|e| e.to_owned());
+                        let example_name = t.example().map(std::borrow::ToOwned::to_owned);
```
dabreegster, line 55: Less readable IMO, but I'm not here to argue with clippy!

osm2lanes-web/src/draw.rs
```diff
@@ -193,20 +117,113 @@ pub fn lanes<R: RenderContext>(
     Ok(())
 }
 
-pub fn draw_arrow<R: RenderContext>(
+// TODO: refactor
+#[allow(clippy::too_many_arguments)]
```
dabreegster, line 121: Honestly the type system saves us here, just imagine this is Python

dabreegster (approved):

osm2lanes-web/src/control.rs
droogmic: yeah, it's strange that `.to_owned` is in scope by default, but `ToOwned::to_owned` is not.

osm2lanes-web/src/draw.rs
droogmic: yeah,  totally workable and maintainable, just a bit of a code smell though.

PR #213, opened 2022-05-25
Add Deserialize to Tags

droogmic:


dabreegster (approved):
This was motivated by the web worker PR, I assume

droogmic:
> This was motivated by the web worker PR, I assume

I noticed it was missing earlier even, when refactoring osm-tags.
But now the webworker really needs this. (or the other refactor)

PR #214, opened 2022-05-29
Gloo Workers v2

droogmic:
This currently fails with:

```
SyntaxError: WorkerGlobalScope.importScripts: Failed to load worker script at "worker.js"
```


droogmic:
This now works, but it depends on an upstream branch I don't control.

osm2lanes/src/test.rs
```diff
@@ -478,6 +478,31 @@ mod tests {
         });
     }
 
+    #[test]
+    fn test_json() {
```
dabreegster, line 482: This was motivated by tracking down where the serde flattening bug was happening, right?

dabreegster (approved):
This appears to work nice and quickly locally, and the code isn't hard to follow! I guess the only concern was `gloo-worker = { git = "https://github.com/futursolo/gloo", rev = "a2101e166260294292c8121fdb8ed883dae62ed8" }`, right? Unless the branch is deleted, there shouldn't be any problems with reproducible builds.

Awesome work!

osm2lanes/src/test.rs
droogmic: yeah, if you replace this with bincode, the test fails :)

PR #216, opened 2022-06-02
Support Cycleway Opposite Track

droogmic:
Required a schema change for markings,
update to the test infrastructure,
a new drawing option

### DIFF
--- a/osm2lanes-web/src/draw.rs
+++ b/osm2lanes-web/src/draw.rs
@@ -1,6 +1,8 @@
 use osm2lanes::locale::Locale;
 use osm2lanes::metric::Metre;
-use osm2lanes::road::{Color as MarkingColor, Designated, Direction, Lane, Printable, Road, Style};
+use osm2lanes::road::{
+    Color as MarkingColor, Designated, Direction, Lane, Printable, Road, Semantic, Style,
+};
 use piet::kurbo::{Line, Point, Rect};
 use piet::{
     Color as PietColor, FontFamily, RenderContext, StrokeStyle, Text, TextAttribute,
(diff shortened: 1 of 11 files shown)


dabreegster (approved):

osm2lanes-web/src/draw.rs
```diff
@@ -107,9 +109,30 @@ pub(crate) fn lanes<R: RenderContext>(
...
+                            scale.scale(left_edge),
+                            0.0,
+                            scale.scale(left_edge + width),
+                            canvas_height,
+                        ),
+                        &PietColor::GREEN,
+                    );
+                    left_edge += width;
+                }
+            },
```
BudgieInWA, line 135: I really like to see this idea. It seems like there should be a default `impl`, and specialisations for different locales. (Would it be worth getting the type system involved in that? Or load from file, like ["schemes" in Map Machine](https://github.com/enzet/map-machine/blob/main/map_machine/scheme/default.yml)?) The different parameters and quirks of the different markings is what gives each style its character and makes it identifiable.

You want to be able to get pretty specific about line drawing, but they can probably be described by a set of cross implementation attributes. Look at the type of thing that the Australian standards (MUTCD type thing) describes:

image

BudgieInWA: What about SVG CSS (in yaml?), as a baseline that can be translated into specialised implementations? There are a bunch of [`stroke` attributes](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute#s) that can be set via css in a browser, and do [all sorts of stuff](https://gigacore.github.io/demos/svg-stroke-dasharray-generator/).

droogmic: I think we need a spec that is based on the ground truth and much simpler? From what I can se SVG is more abstract.

All I need is a width in metres, and a `Vec<Metre>` to represent the repeating pattern?

Do we even have any examples where there is a repeating pattern in road paint?

dabreegster: Related conversation in https://github.com/a-b-street/osm2streets/issues/24

PR #219, opened 2022-06-09
OSM Tag Schemes

droogmic:
Apologies, this PR suffered from scope creep :/


osm-tag-schemes/src/highway.rs
```diff
@@ -228,6 +208,24 @@ impl Highway {
         }
     }
 
+    /// Active Highway
```
droogmic, line 211: construction

osm-tag-schemes/src/highway.rs
```diff
@@ -228,6 +208,24 @@ impl Highway {
...
+    #[must_use]
+    pub fn construction(r#type: HighwayType) -> Self {
+        Self {
+            highway: r#type,
+            lifecycle: Lifecycle::Construction,
+        }
+    }
+
+    /// Active Highway
```
droogmic, line 220: proposed

osm2lanes/src/tag_keys.rs
```diff
@@ -0,0 +1 @@
+
```
droogmic, line 1: delete

BudgieInWA:
Sometimes a band-aid needs to be ripped off, and the changes need to be made together :P 

This looks exciting! What are your goals here?

droogmic:
This extends your design of having a semantic tagging scheme exist as an intermediate step, that can be useful for more than just our client.

If `Tags` exists in a basic crate, then we build another crate on top of it that does tagging scheme parsing, and then we put osm2lanes on top of both :)

BudgieInWA:
Great! A description of this setup in the README (under structure?) would be awesome.

I agree that the description of individual tagging schemes are complicated enough on their own, and should exist independently like that. It feels like a place to collect together the documentation for what all the variants mean, how they should be interpreted, (links to examples?), references to the wiki and usage, ...

It is like they ought to extend / merge with the wiki pages [(e.g.)](https://wiki.openstreetmap.org/wiki/Tag:cycleway=lane).

PR #221, opened 2022-06-10
Cleanup TagsToLanesMsg Interfaces

droogmic:


osm2lanes/src/transform/tags_to_lanes/error.rs
```diff
@@ -95,11 +95,11 @@ impl TagsToLanesMsg {
 
     #[must_use]
     #[track_caller]
-    pub fn unsupported(description: &str, tags: Tags) -> Self {
+    pub fn unsupported(description: String, tags: Tags) -> Self {
```
dabreegster, line 98: This PR makes all the call-sites have more boilerplate. How about using `<T: Into<String>>` here? Slightly slower compilation from the generics, but the code elsewhere stays a bit more legible

PR #224, opened 2022-06-12
Deserializing Tags Crashing on Website

droogmic:
It appears to be due to a duplication in tags in overpass' JSON response.

e.g. given way 3600044874
ISO3166-2=IT-21 is repeated 4 times.
The cause is unclear.


dabreegster (approved):
That's what we get for being paranoid about input, I guess

PR #226, opened 2022-06-14
Crate Deploy

droogmic:


.github/workflows/crate-deploy.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: "publish osm2lanes"
+        run: |
+          rustup toolchain install stable
```
dabreegster, line 15: Do we not need `hecrj/setup-rust-action@v1`? How is `rustup` already around?

droogmic: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#rust-tools

.github/workflows/crate-deploy.yml
```diff
@@ -0,0 +1,18 @@
...
+          rustup override set stable
+          cd osm2lanes/
+          cargo publish --dry-run
```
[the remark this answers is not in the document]
droogmic: So I think this is fine as a workflow

.github/workflows/crate-deploy.yml
dabreegster: Woo, nice!

PR #229, opened 2022-06-18
Refactor and Prepare to add Neukolln Example

droogmic:


data/tests.yml
```diff
@@ -803,6 +803,52 @@
...
+    highway: "secondary"
+    lanes: "3"
+    lanes:backward: "2"
+    lanes:forward: "1"
+    sidewalk:both: "separate"
+    bicycle:lanes:backward: "yes|designated|yes"
+    cycleway:both: "lane"
+    cycleway:both:surface:colour: "red"
+    cycleway:lanes:backward: "no|lane|no"
+    cycleway:left:width: "1.5"
```
dabreegster, line 819: Exciting to look forward to!

osm-tags/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "osm-tags"
-version = "0.1.0"
+version = "0.1.1"
```
dabreegster, line 3: So if I understand the github workflow, it'll try to publish this crate every single time we change `osm-tags`, but only work if we increment this version number? So we can more or less decide when we want to push a new version

dabreegster (approved):
No noteworthy changes in the code moved around, I'm assuming

osm-tags/Cargo.toml
droogmic: yeah, its the simplest thing that works, to avoid needing to discuss "branching and release strategies".

2 additional tweaks to add later:
- disable it when there are no changes, so that actions doesn't report a failure
- require a version bump if there are changes, in the test workflow

droogmic:
> No noteworthy changes in the code moved around, I'm assuming

the only semantic change was in some of the error paths regarding access by lane, in tests I trust.

PR #235, opened 2022-06-18
Add Circle When Searching

droogmic:
Visual indication of loading.

### DIFF
--- a/osm2lanes-web/src/map.rs
+++ b/osm2lanes-web/src/map.rs
@@ -188,7 +188,6 @@ impl MapComponent {
         if let Some(search_circle) = self.search_circle.take() {
             search_circle.remove();
         }
-        self.search_circle = None;
     }
 }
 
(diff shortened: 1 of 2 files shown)


osm2lanes-web/src/map.rs
```diff
@@ -161,6 +174,24 @@ impl Component for MapComponent {
...
+        );
+        search_circle.addTo(&self.map);
+        self.search_circle = Some(search_circle);
+    }
+
+    fn stop_search(&mut self) {
+        if let Some(search_circle) = self.search_circle.take() {
+            search_circle.remove();
+        }
+        self.search_circle = None;
```
dabreegster, line 191: This is redundant with `take` above

dabreegster (approved):
Awesome UX improvement!

osm2lanes-web/src/map.rs
droogmic:
```suggestion
```

PR #236, opened 2022-06-19
Fetch Geometry

droogmic:


dabreegster (approved):
The plumbing looks good to me. Running locally, I can't tell what effect loading the geometry has. If you type in a way ID, the map doesn't autozoom there and show it or anything like that. That's not meant to happen, right?

droogmic:
> The plumbing looks good to me. Running locally, I can't tell what effect loading the geometry has. If you type in a way ID, the map doesn't autozoom there and show it or anything like that. That's not meant to happen, right?

hmm, so I guess it was in draft for a reason then, sort of forgot what I was doing... :P

Issue #240, opened 2022-07-22
Web app to edit lanes and upstream changes in OSM


dabreegster:
I have now started a very minimal prototype in https://github.com/a-b-street/osm2lanes/tree/web_editor to make a new web app that lets people select an OSM way, see the rendered lanes in a cross-section view, edit those lanes (adding, deleting, reordering, modifying width, etc), generate the OSM tags for the changes, and upload to OSM. This is a step towards making it easier to edit lane data in OSM.

@BudgieInWA and I discussed some of the complications that'll come up. Two of the big ones:

1) When a road is split into multiple parallel ways, what do we do? The lane editor as I'm building it operates on a single way. We could maybe ask the user to draw a line and select all of the ways they want to edit, then present a unified cross section across all of them, and make it very clear which way is getting edited. Of course it won't be that simple, so to start, my goal is just to handle simple cases where one way corresponds to the "entire" road.

2) The left-to-right order of the draggable lane cards of course depends how the way in OSM is oriented. We need to display the way on the map and make it clear what's happening. Even better, leverage osm2streets + some rendering to draw the lanes on the 2D map at the same time.

I'm building using vanilla JS, which is new to me (this is a learning project!).

When I get something minimally decent open, I'll open a PR -- hopefully a few days if I can keep dodging other responsibilities. :)

dabreegster:
@BudgieInWA had asked by email:

> In the course of implementing the editing capabilities, it seems like you will be writing `reorder_lanes(...)`, but also `add_lane_at_position` and `remove_lane_at_position` etc., even `add_sidewalk`, `add_bus_lanes`, etc. It occurs to me that those logical modifications to an existing road are the same sorts of modifications that osm2lanes wants to make during the "building" stage of the process. I wonder if one of the goals of osm2lanes should be a mutable `Road` type with an API designed for editor applications s to use interactively. That way, the current state (list of lanes) can be returned from wasm to JS, the current state can be rendered based on it, then the editor can capture the intent to make some change that is supported -- such as `reorder`, `remove_lane`, `add_sidewalk` -- which can be sent back to wasm, to let osm2lanes figure out the nuances of that change, before returning the new state to update the UI accordingly.

[...]

PR #241, opened 2022-07-25
Start a web app to edit lanes and upstream tags to OSM

dabreegster:
#240 
The goal for this app is to make it easier to tag lanes correctly in OSM. Somebody would pick a way, see the current lanes in a cross-section view, drag the cards around and make things look correct, generate OSM tags to represent that, and upload the changeset.

This PR is a minimal prototype of that flow, without uploading changesets yet. I'd like to make sure the basic structure is OK in code review, then send followups to flesh out its functionality more. I'm giving myself an aggressive goal of mid August to make this actually usable end-to-end in simple situations, for the SoTM conference. It's a non-goal right now to handle roads split into multiple ways.

To run it, you should just be able to `cd web_editor; ./serve_locally.sh`. The only dependency is `wasm-pack` to compile and `python3` to serve a dummy HTTP file server. The demo is not very impressive yet:

https://user-images.githubusercontent.com/1664407/180864376-6f4b2984-8e62-4f5a-a8c4-f5bfa671b7a3.mp4


web_editor/index.html
```diff
@@ -0,0 +1,66 @@
...
+          border: 2px dashed red;
+          min-height: 100px;
+          padding: 30px;
+        "
+      >
+        Delete lane
+      </div>
+
+      <div id="toolbox" class="cards-row" style="float: left">
+        <!-- TODO Can I just make up / overload the value attribute here? -->
```
dabreegster, line 50: I'm just sort of stuffing an "enum" in here as `value`, then later in the JS turning this into JSON data for a lane. Is that a half-reasonable pattern to start with?

web_editor/js/main.js
```diff
@@ -0,0 +1,147 @@
...
+// https://github.com/thedodd/trunk/issues/230 or stop using trunk.
+import { dummyData } from "./dummy_data.js";
+import init, {
+  js_way_to_lanes,
+  js_lanes_to_tags,
+} from "./osm2lanes-npm/osm2lanes_npm.js";
+await init();
+
+// Setup the toolbox controls
+new Sortable(document.getElementById("toolbox"), {
```
dabreegster, line 11: I'm using http://sortablejs.github.io/Sortable/ for the drag-and-drop functionality. It's pretty good so far, though I haven't figured out how to do some things with it. If anyone's used another library that I should look at, lmk

web_editor/js/main.js
```diff
@@ -0,0 +1,147 @@
...
+// TODO Disable the ghostClass of lanes when going here
+new Sortable(document.getElementById("delete"), {
+  group: "lanes",
+  onAdd: function (evt) {
+    var el = evt.item;
+    el.parentNode.removeChild(el);
+  },
+});
+
+// This state is set up by start_editing
```
dabreegster, line 30: The state management is partly here. These vars represent the original version of the currently fetched way. Then the "current" version of the lanes exists in the DOM, within the ordered `cards` div. Each card has a `laneJSON` attribute smushed into it. I think this lets us avoid any hassle of syncing the DOM with state kept elsewhere. Is this a reasonable pattern?

web_editor/js/main.js
```diff
@@ -0,0 +1,147 @@
...
+var current_road = null;
+var current_locale = null;
+
+async function start_editing() {
+  const way = BigInt(document.getElementById("osm_way_id").value);
+  // Faster dev workflow: if the way ID is the default, use baked-in data instead of waiting on Overpass.
+  var road_wrapper, locale;
```
dabreegster, line 37: I want to just do `const [x, y] = if (blah) { x } else { y }`...

web_editor/js/main.js
```diff
@@ -0,0 +1,147 @@
...
+  if (way == 804788513) {
+    [road_wrapper, locale] = dummyData();
+  } else {
+    console.log(`Fetching ${way}...`);
+    [road_wrapper, locale] = await js_way_to_lanes(way);
+  }
+  const road = road_wrapper["Ok"]["road"];
+  console.log(`Got osm2lanes output, creating cards`);
+
+  current_road = JSON.parse(JSON.stringify(road));
```
dabreegster, line 47: Is there a more canonical way to deep clone?

web_editor/js/main.js
```diff
@@ -0,0 +1,147 @@
...
+  current_locale = JSON.parse(JSON.stringify(locale));
+
+  // TODO Fully clean up the old cards, including whatever sortable thing is attached there?
+  const cards = document.getElementById("cards");
+  cards.replaceChildren();
+
+  // Create a card per lane
+  for (const lane of road["lanes"]) {
+    if (lane["type"] == "separator") {
```
dabreegster, line 56: I'll probably hash this out in the next PR, but I might make non-draggable cards for all of the separators. It's at least important to show the center line / where directions change. Then when you drag any card across that line, we should probably flip the direction automatically.

web_editor/js/main.js
```diff
@@ -0,0 +1,147 @@
...
+          type: "parking",
+          direction: "backward",
+          designated: "motor_vehicle",
+        });
+      }
+      evt.item.replaceWith(card);
+    },
+  });
+}
+window.start_editing = start_editing;
```
dabreegster, line 97: Or we could `document.getElementById("Edit").on_click = start_editing`. Is there a style I should prefer?

web_editor/js/main.js
```diff
@@ -0,0 +1,147 @@
...
+
+function make_lane_card(lane) {
```
dabreegster, line 99: Zero attempt at style yet, obviously

dabreegster:
@michaelkirk, I'm starting some vanilla JS + WASM projects. Any interest / time to help review?

droogmic (approved):
I guess it would be too hard to somehow plug this into the streetmix source code

dabreegster:
> I guess it would be too hard to somehow plug this into the streetmix source code

From a quick skim through, it wouldn't be easy! But also their UI isn't designed for data entry about existing reality, and their level of detail is way higher than OSM's on street furniture, scooter parking zones, etc. I'm planning to adapt some of their design choices (draggable cards, cross section view, width adjustments, etc)

droogmic:
I can't do a detailed review, because I am allergic to JS (partial sarcasm).

web_editor/js/main.js
michaelkirk: how about code style? 😬 

```suggestion
function makeLaneCard(lane) {
```

It's not unheard of for people to write javascript like it's c, but it's not typical. I suppose it ultimately depends if you want this web code to feel more familiar to rust programmers or to web programmers.

michaelkirk (approved):
I had some suggestions for application structure, but nothing that I'd consider blocking.

Nice to see you delving into web-native UI!

web_editor/js/main.js
dabreegster: Closer to the second -- they're a concept in the `osm2lanes` API representing some kind of semantic or physical separator between lanes:
Screenshot from 2022-07-27 12-12-38
In order to keep the osm2lanes JSON state in sync with the DOM (managed by sortable), I may insert invisible or not-movable cards to represent the separators. Not sure yet.

web_editor/js/main.js
dabreegster: Thanks for the suggestion! I started a class to bundle all of the state together, and it's much more comfortable. I'm still a bit confused about scoping in separate .js files and how to reference things from HTML, but I'll figure out the patterns over time.

> But maybe the eventual solution to dual carriage ways will be that we'll have two instances of this editor thing going at once 

Spot on -- I think Ben has brought that idea up. And we may also want to architect the logic in a way that could eventually be directly part of the iD editor, so bundling state together is smart in any case. Thanks for the explanation!

web_editor/js/main.js
dabreegster: Oops, fixed. I have to un-Rust my brain, or find a linter that yells at me.

dabreegster:
Thanks for the reviews! I'll merge this initial prototype in and open more PRs as this takes real shape.

PR #242, opened 2022-07-29
remove some global var usage

michaelkirk:


web_editor/index.html
```diff
@@ -18,14 +18,17 @@
       }
     </style>
 
-    <script type="module" src="js/main.js"></script>
+    <script type="module">
+        import { LaneEditor } from "./js/main.js"
+        window.editor = await LaneEditor.create();
```
dabreegster, line 23: Ah, so this is how to decouple where the `LaneEditor` is stored. Thank you, I understand the pattern (and how to export stuff from modules) now!

dabreegster (approved):
Thanks!

PR #253, opened 2022-11-01
Fully switch over to new test output

dabreegster:
The expected test output used to just be a list of lanes. When we added the outer `Road` structure, some of the tests changed, but not all. This PR fully migrates to the new expected output style and cleans up the transition stuff.

Overall goal: simplify how the tests work, make it easy to crank out new ones.


dabreegster:
I have high confidence this change is correct and uncontroversial, so merging

PR #256, opened 2022-11-01
Add a button to generate example test YAML from the web app

dabreegster:
If you want to add a test case to the .yaml today, you have to:

- copy paste an existing example, then change some things about it
- decide where in the .yaml file it should go (based on comments about organization)

This PR mostly fixes the first. You can use the web app, get the current output, and generate the YAML. You have to manually fix it up -- usually because the expected output is wrong -- but this is much faster workflow.
Screenshot from 2022-11-01 11-58-35

### DIFF
--- a/osm2lanes-web/Cargo.toml
+++ b/osm2lanes-web/Cargo.toml
@@ -32,6 +32,7 @@ wasm-bindgen = "0.2"
 wee_alloc = { version = "0.4", optional = true }
 yew = "0.19"
 gloo-worker = { git = "https://github.com/futursolo/gloo", rev = "a2101e166260294292c8121fdb8ed883dae62ed8" }
+serde_yaml = "0.9"
 
 [dependencies.web-sys]
 version = "0.3"
--- a/osm2lanes-web/src/lib.rs
+++ b/osm2lanes-web/src/lib.rs
@@ -389,7 +389,7 @@ fn generate_test_yaml(
     locale: &Locale,
     way_id: Option<i64>,
 ) -> String {
-    use osm2lanes::test::{Expected, TestCase};
+    use osm2lanes::test::TestCase;
 
     let test = TestCase {
         way_id,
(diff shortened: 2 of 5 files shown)


osm2lanes-web/Cargo.toml
```diff
@@ -32,6 +32,7 @@ wasm-bindgen = "0.2"
 wee_alloc = { version = "0.4", optional = true }
 yew = "0.19"
 gloo-worker = { git = "https://github.com/futursolo/gloo", rev = "a2101e166260294292c8121fdb8ed883dae62ed8" }
+serde_yaml = "0.9"
```
dabreegster, line 35: https://github.com/dtolnay/serde-yaml/issues/87
Without this, we get `field: ~` in the output. The version bump gets us to `field: null`, which is still annoying, and we strip it out manually

osm2lanes-web/src/lib.rs
```diff
@@ -363,3 +376,34 @@ impl App {
...
+        expected: Expected::Road(road.clone()),
+        rust: None,
+    };
+    // TODO Strip out road's name, ref, and other things we don't test for?
+    // TODO Based on checkboxes, strip out separators
+    let raw = serde_yaml::to_string(&test).unwrap();
+
+    // serde_yaml explicitly lists "field: null" for None values. Filter these out, to match the
+    // style of the test YAML.
+    let mut output = String::new();
```
dabreegster, line 401: The output isn't indented properly for copying into the .yaml, and I'm having trouble getting this working... will iterate later

dabreegster:
I was going to add a "copy to clipboard" button for convenience, but it's unexpectedly tough in Yew.

Option 1: just have an `onclick="someJSFunction()"`. I can't get this to work; The yew HTML macro insists on callbacks being in Rust, and the docs don't talk about any escape hatch to just call JS. I could define the function in JS and make a WASM interface to it, I guess. https://yew.rs/docs/0.18.0/concepts/html/elements#listeners

Option 2: Copy to the clipboard via rust. But the necessary APIs are unstable. https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.Clipboard.html

So for now, this tool is helpful and I'll merge. Will still add more options to filter out separators / widths later.

PR #258, opened 2022-11-02
Flesh out the web editor, remove drag-and-drop

dabreegster:
https://user-images.githubusercontent.com/1664407/199611654-8e9d39aa-a323-4a4e-b285-5abaf9fb0d82.mp4

Remove drag-and-drop for now, since it complicates the code, we were using a library I'm not sure is the right choice, and the UX of dragging a trash can to a lane vs a lane to a trash can is a bit unclear to me. Instead, just have some buttons inline to each card, and add new lanes to the end always.


web_editor/js/main.js
```diff
@@ -6,17 +6,32 @@ import init, {
...
+  constructor(way, road, locale, tags) {
     this.way = way;
-    // Clone these
-    this.currentRoad = JSON.parse(JSON.stringify(road));
-    this.currentLocale = JSON.parse(JSON.stringify(locale));
+    this.road = road;
+    this.locale = locale;
     this.originalTags = tags;
+
+    // Remove these immediately. If they're rendered invisibly, they mess up indices.
```
dabreegster, line 25: Still temporary. Seeing the center line where direction changes would be quite helpful visually.

web_editor/js/main.js
```diff
@@ -6,17 +6,32 @@ import init, {
...
+    this.road.lanes = this.road.lanes.filter(
+      (lane) => lane.type != "separator"
+    );
+
+    // All of the state is immutable, except for road. The source of truth for
```
dabreegster, line 30: This is different -- and simpler -- than before, when the source of truth was the sortable cards in the DOM. Anytime we make a mutation, we just re-render all of the cards from scratch. This is cheap enough.

web_editor/js/main.js
```diff
@@ -112,44 +85,47 @@ export class LaneEditor {
...
+        this.#diffTags();
       } catch (err) {
         window.alert(`Error: ${err}`);
       }
-    }
-  };
+    };
+
+    document.getElementById("new-driving").onclick = () => {
+      this.road.lanes.push({
```
dabreegster, line 99: Always appending to the end is clunky. You had suggested having + buttons in between each pair of lanes, and presumbly a way to click the type and change it. I like that idea now that I type it out; I'll try that in a followup

Robinlovelace:
This looks as easy to user as streetmix.net, if not more so. And it's simpler which is good (and more relevant to OSM). Transport focused OSM editing functionality here we come?

michaelkirk:
Sorry I haven't been following development here that closely, so feel free to ignore me if I'm off base.

I felt pretty good about the UI we iterated on for the abstreet lane editor, where you could drag-and-drop lanes to re-order them, but deleting was a click-button action. 

Maybe that's a reasonable route to work towards for this web implementation:

https://user-images.githubusercontent.com/217057/199624454-1bbce7b6-11fb-4e80-a514-94d3d26575a6.mp4

dabreegster:
> I felt pretty good about the UI we iterated on for the abstreet lane editor, where you could drag-and-drop lanes to re-order them, but deleting was a click-button action.

Good feedback, thanks! I think the A/B Street behavior is what we want to build towards here. This JS editor died off for a few months, and I want to get the ball rolling again by attempting https://github.com/a-b-street/osm2lanes/issues/240#issuecomment-1288582528. Drag-and-drop will come back, but maybe not with sortable.js. And having a separate button for deletion (instead of dragging a lane <-> a trash can) sounds like simpler UX too.

web_editor/js/main.js
BudgieInWA: The simplicity gained by this approach is a huge win in my experience.

BudgieInWA (approved):
This is a mobile phone review, so I'm yet to run or play with the app, but the videos look great and the code is easy to follow. I've just added a couple of low-level comments.

web_editor/js/cards.js
```diff
@@ -61,3 +94,9 @@ function width(lane) {
 function icon(name) {
   return `<img src="assets/${name}.svg" class="icon" />`;
 }
+function iconObj(name) {
+  var obj = document.createElement("img");
+  obj.src = `assets/${name}.svg`;
+  obj.className = "clickable";
```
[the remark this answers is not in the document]
dabreegster: Great point! Fixed

dabreegster:
PTAL, I switched over to always building up DOM elements, and using buttons.

web_editor/js/cards.js
```diff
@@ -1,13 +1,47 @@
...
-  node.laneJSON = lane;
+  node.className = "card";
+  node.title = JSON.stringify(lane, null, 2);
+  node.style = `background: ${backgroundColor(lane)};`;
 
-  node.innerHTML = `<div align="center">` + typeIcon(lane) + `</div>`;
-  node.innerHTML += `<div align="center">` + directionIcon(lane) + `</div>`;
-  node.innerHTML += `<div align="center">` + width(lane) + `</div>`;
+  node.innerHTML = `<div align="center">${typeIcon(lane)}</div>`;
+  node.innerHTML += `<div align="center">${directionIcon(lane)}</div>`;
```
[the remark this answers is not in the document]
BudgieInWA: *Assigning* to `innerHTML` is fine, it's just the *append* that is problematic, as described in this answer: https://stackoverflow.com/a/23339072

I think the template literals approach is a good style, just commit it to the DOM in one go :)

web_editor/main.css
```diff
@@ -8,36 +8,28 @@
...
 }
 
+.card button {
+  background: none;
+  border: none;
+}
+.card button:hover:enabled {
+  opacity: 50%;
+}
+
```
BudgieInWA, line 18:
```suggestion
.card button:disabled {
  opacity: 25%;
}

```

BudgieInWA (commented):
I got it running locally, and it's looking good!

web_editor/js/cards.js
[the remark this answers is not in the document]
dabreegster: Ahhhh thank you, I understand now. Serializing to HTML and then parsing everything out again sounds awful!

I'll keep things as all objects for now, because the next step will need to make the direction arrow clickable. AFAICT, there's not a nice way to mix template literals and attaching callbacks, without assigning an ID and then looking it up, or [making one callback / listener at the parent level](https://stackoverflow.com/a/31876859)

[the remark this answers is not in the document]
BudgieInWA: Using listeners on some parent can work well, especially using [`data-` attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) instead of `id`s to communicate data.

PR #260, opened 2022-11-07
More edit controls

dabreegster:
#240 

https://user-images.githubusercontent.com/1664407/200313141-793b6aff-1e2f-4bf3-9048-4be609614850.mp4

1) A (+) button to insert a lane at a particular position
2) Add missing card for sidewalks!
3) Click direction arrow to toggle it

A next possible step is clicking the lane type icon to get a little modal popup and choose the type. So then inserting a lane and changing the type would become much more fluid


web_editor/main.css
```diff
@@ -33,6 +35,13 @@
   width: 50px;
 }
 
+.insert-lane {
+  position: absolute;
+  right: -15px;
```
dabreegster, line 40: Screenshot from 2022-11-07 11-50-38
Actually, I got the icons to be in a better place. Is this a reasonable way of achieving it?

BudgieInWA: Perfect use of `position: absolute`! We want to anchor it to the bottom right of the card and that's idiomatic CSS to do so.

web_editor/js/cards.js
```diff
@@ -11,7 +11,8 @@ export function makeLaneCard(lane, idx, app) {
   }
   node.appendChild(wrapInCenterDiv(width(lane)));
 
-  var finalRow = document.createElement("div");
+  var editRow = document.createElement("div");
+  editRow.align = "center";
```
BudgieInWA, line 15: I'd prefer this done in CSS. `text-align: center` (for centering inline children of a block element) or `margin: 0 auto` (for centering a block element within its parent) will cover most cases.

BudgieInWA (approved):
:+1:

Using a "dropdown" type interface for changing lane type as you describe sounds good.

web_editor/js/cards.js
dabreegster: Done!

PR #261, opened 2022-11-08
Create an offline geocoder to determine driving side from a coordinate.

dabreegster:
Part of https://github.com/a-b-street/osm2streets/issues/80. Today we require the user to specify the driving side, or we attempt to query it from Overpass (but that seems to be flaky). It's much simpler to just see if some point within an imported area is within the polygon of a left-sided country or not.

See README.md for data sources and inspiration. I've got this tested end-to-end in osm2streets; a PR for that will be coming momentarily. I think we can later cleanup more things in osm2lanes based on this.

I'd like to later split out `driving-side` into its own repo and put it on crates.io, since it's of standalone value. But first I'd generalize it and return the ISO country code, not just left/right driving side -- aka, make a Rust port of https://github.com/westnordost/countryboundaries. I'd also add benchmarks and try out some optimizations. But for now, `driving-side` is fast enough and has immediate massive value to osm2streets and A/B Street.


driving-side/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+use geo::{Contains, Point};
+use geojson::{quick_collection, GeoJson};
+
+/// Returns true if the point is located in a country with left-handed driving.
+pub fn is_left_handed(pt: Point) -> bool {
+    let raw = include_str!("../data.geojson");
```
dabreegster, line 6: Every single query does the work of parsing the geojson -- surely that's slow? In practice, not at all -- `cargo test` in debug mode is under 0.01s for the 4 cases. And the usage pattern of this API is once per map import, so the time is absolutely negligible.

We could optimize later by
1) making an API to do the loading+parsing once, then query many times
2) using a quadtree like https://github.com/georust/rstar to prune the lookup
3) Serializing the quadtree and `geo::Polygon`s directly in a binary format and deserializing that instead

dabreegster: Oh yeah, and including this crate adds about 80KB to build size from `data.geojson`. That's also negligible and could be made smaller by step 3 above

dabreegster:
Feel free anyone to review asynchronously. To avoid PRs in a few repos stacked on each other and to make progress on the original bug I was working on (yak-shaving is fun when it's this easy and beneficial!), I'm going to merge now

PR #266, opened 2023-02-02
Generalize the driving-side crate to return ISO country codes

dabreegster:
This adds all countries, the iso2 code, and cleans up the API slightly. As a next step, I'll split into a separate repo and publish to crates.io.

The intention is that we can use this in osm2streets to start defining `typical_lane_widths` that vary by country code.


dabreegster (commented):
The bundled file is now 400KB. The tests are still extremely subsecond. Plenty of perf ideas if they ever matter (topojson, storing binary serialized serde stuff, using a quadtree to speed the lookup)

dabreegster:
Going to merge this and carry forward with splitting into a new repo, and using from osm2streets

